PR #62, opened 2016-10-09
Publisher app

christophehenry:


jaywink:
As per IRC talk I'll close this. No need for new publisher app, I think the publisher changes you want to make fit nicely into the existing `content` app.

PR #63, opened 2016-10-15
Bootstrap-markdown on publisher

christophehenry:
screen shot 2016-10-15 at 10 33 50

**Note:** The boostrap-markdown plugin is not compatible with BS4 yet. So I need an upstream PR to be merged to fix the buttons. Also, some FontAwesome icon seems to not work properly.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -71,6 +71,7 @@ module.exports = function (grunt) {
                     '<%= paths.css %>/common.css': '<%= paths.sass %>/common.scss',
                     '<%= paths.css %>/grids.css': '<%= paths.sass %>/grids.scss',
                     '<%= paths.css %>/streams.css': '<%= paths.sass %>/streams.scss',
+                    '<%= paths.css %>/publisher.css': '<%= paths.sass %>/publisher.scss'
                 },
             },
             dist: {
@@ -84,6 +85,7 @@ module.exports = function (grunt) {
                     '<%= paths.css %>/common.css': '<%= paths.sass %>/common.scss',
                     '<%= paths.css %>/grids.css': '<%= paths.sass %>/grids.scss',
                     '<%= paths.css %>/streams.css': '<%= paths.sass %>/streams.scss',
+                    '<%= paths.css %>/publisher.css': '<%= paths.sass %>/publisher.scss',
                 },
             }
         },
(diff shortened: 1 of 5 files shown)


jaywink:
Thanks for working on this. Some early stage comments:
- Would be nicer to have the preview automatic, on the side on large monitors and underneath on smaller devices. That is how the current editor works. That would eliminate need for preview button also. Of course it needs to be rate limited so when the user is typing it only fires after the typing pauses.
- Dragging an image into the publisher should POST it via Ajax to some image saving view which returns the url for it - just like the current markdownx does, so probably good idea to look into code there.
- The preview doesn't seem to hook onto the backend at the moment? We want to keep the markdown rendering in the backend (for optimization), so instead of using bs-markdown internal rendering, some hook to an Ajax call is needed for the preview.

christophehenry:
Ok, so that's even easier. I just keep the current markdownx plugin. Just need to change some things a bit.

christophehenry:
Here it is:

screen shot 2016-10-29 at 12 42 43

I have just a tiny SCSS problem: I cant use BS' variables and mixins. There's probably something to fix in the asset processing.

jaywink:
selection_113

Pulled in bower installs, please apply this change:

```
diff --git a/bower.json b/bower.json
index abb1809..2b69952 100644
--- a/bower.json
+++ b/bower.json
@@ -28,5 +28,8 @@
   "devDependencies": {
     "sinon": "http://sinonjs.org/releases/sinon-1.17.4.js",
     "mock-socket": "^2.0.2"
+  },
+  "resolutions": {
+    "bootstrap": "v4.0.0-alpha.5"
   }
 }
```

christophehenry:
Done.

jaywink:
Really don't understand why bower asks, I mean it's specified directly :)

I'm not sure what goes with the bs variables - I was wondering the same at some point. Need to investigate closer, probably some `grunt-sass` config thingy or bug.

Two minor things really only:
- The preview could do with falling underneath already on "bs small" ie 544. Now it just ends up halving it's width, in the end falling underneath when screen gets smaller. See giffy attached.
- Seems the header broke probably due to new bootstrap alpha version. I can fix it later so don't worry too much unless you have a quick idea about it.

peek 2016-10-30 20-24

Otherwise everything looks really good and awesome, nice work! <3 If the small glitch on preview scaling can't easily be fixed I can merge this as is.

christophehenry:
The current bootstrap-markdown plugin is not ready fot BS4. I have a currently opened PR upstream to let user set custom classes to the buttons so we can fix it when it's merged which can get a long time :)

jaywink:
The buttons look OK to me? In the gif I made.

christophehenry:
Done.

jaywink:
Awesome, it's merging time then! Works perfectly for me on two environments, can't see anything wrong with the icons. A big thanks for this - you're the first contributor to Socialhome ever! ✨ ❤️ So if this becomes a super big successful project you're basically a history maker! 🏄

christophehenry:
\o/
About the buttons, they use your system's configuration. That's why they are grey with gold text on the picture of my first post. But now that [bootstrap-markdown#262](https://github.com/toopay/bootstrap-markdown/pull/264) is merged, we can use BS4's classes until BSMD is ported to BS4.

Issue #61, opened 2016-10-09
Render logged in user profile on root page


jaywink:
Even for single user mode where landing page is a set profile, when a user is logged in, always render them their own profile on root.

The relevant place to do the redirect would be somewhere [around here](https://github.com/jaywink/socialhome/blob/9bcd575bd6dc4879dcaed903fdb6d348a99b937a/socialhome/content/views.py#L78-L78).

masterp4dev:
Hi @jaywink, I would like to work on this. Any heads up?

jaywink:
Hey @masterp4dev - awesome!

Basically Socialhome works in two modes - "root page mode" or "root profile mode". If a root profile is set, instead of a generic root page (`HomeView`) a set user profile page is rendered (`ProfileDetailView`). You can see this happening [here](https://github.com/jaywink/socialhome/blob/9bcd575bd6dc4879dcaed903fdb6d348a99b937a/socialhome/content/views.py#L78-L78).

What we want to change is to redirect to the **logged in user profile page** (`ProfileDetailView`) if a user is logged, always. So that has to override both the possibilities above.

The readme has instructions on how to set up an environment. If you have any question please don't hesitate to ask here or on IRC FreeNode on channel #socialhome (http://webchat.freenode.net?channels=%23socialhome&uio=d4).

christophehenry:
Can't move the card :'(

jaywink:
Closed via #105

PR #105, opened 2017-03-04
Render logged in user profile on root page

christophehenry:
#61

### DIFF
--- a/socialhome/content/tests/test_views.py
+++ b/socialhome/content/tests/test_views.py
@@ -10,17 +10,29 @@ from socialhome.users.models import Profile
 from socialhome.users.tests.factories import UserFactory
 
 
-@pytest.mark.usefixtures("admin_client", "settings")
+@pytest.mark.usefixtures("db")
 class TestRootProfile(object):
-    def test_home_view_rendered_without_root_profile(self, admin_client, settings):
+    @pytest.mark.usefixtures("client", "settings")
+    def test_home_view_rendered_without_root_profile(self, client, settings):
         settings.SOCIALHOME_ROOT_PROFILE = None
-        response = admin_client.get("/")
+        response = client.get("/")
         assert response.templates[0].name == "pages/home.html"
 
-    def test_home_view_rendered_with_root_profile(self, admin_client, settings):
-        settings.SOCIALHOME_ROOT_PROFILE = "admin"
+    @pytest.mark.usefixtures("admin_client", "settings")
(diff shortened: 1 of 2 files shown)


christophehenry:
Halp! :(

jaywink:
You need to fix the tests to match the new functionality. Test `test_home_view_rendered_without_root_profile` in https://travis-ci.org/jaywink/socialhome/jobs/207769585#L734 is checking that `home.html` is rendered, but now `profile.html` is rendered, as the test is using a logged in user.

These test adjustments might work? There is a new test and also checking that correct profile is in context. I want to test everything in Socialhome, so you'll need to learn py.test and Django testing.. :)

    @pytest.mark.usefixtures("db")
    class TestRootProfile(object):
        @pytest.mark.usefixtures("client", "settings")
        def test_home_view_rendered_without_root_profile(self, client, settings):
            settings.SOCIALHOME_ROOT_PROFILE = None
            response = client.get("/")
            assert response.templates[0].name == "pages/home.html"
    
        @pytest.mark.usefixtures("admin_client", "settings")
        def test_logged_in_profile_view_rendered_without_root_profile(self, admin_client, settings):
            settings.SOCIALHOME_ROOT_PROFILE = None
            response = admin_client.get("/")
            assert response.templates[0].name == "streams/profile.html"
            assert response.context["profile"].user.username == "admin"
    
        @pytest.mark.usefixtures("client", "admin_client", "settings")
        def test_home_view_rendered_with_root_profile(self, client, admin_client, settings):
            settings.SOCIALHOME_ROOT_PROFILE = "admin"
            # Set admin profile visibility, otherwise it will just redirect to login
            Profile.objects.filter(user__username="admin").update(visibility=Visibility.PUBLIC)
            response = client.get("/")
            assert response.templates[0].name == "streams/profile.html"
            assert response.context["profile"].user.username == "admin"

christophehenry:
Your adjusment work. Though I do not clearly understand why, yet.

socialhome/content/tests/test_views.py
```diff
@@ -1,3 +1,5 @@
+from types import SimpleNamespace
```
jaywink: This import is unused now I think?

jaywink (changes requested):

socialhome/content/tests/test_views.py
christophehenry: Yes, it is. I forgot that one.

jaywink (approved):
Awesome, thanks!

Issue #106, opened 2017-03-05
Automatically order new pinned profile content as last


jaywink:
Currently when content is created and it is pinned to the profile, it gets inserted with a default ordering, which means it is likely near the top.

Make new content have ordering as "current max +1" so that new pinned content is shown last.

jaywink:
Closed via #107

PR #107, opened 2017-03-06
Automatically order new pinned profile content as last (#106)

christophehenry:


socialhome/content/models.py
```diff
@@ -118,6 +118,8 @@ def save(self, author=None, *args, **kwargs):
             # New with author, set a GUID and author
             self.guid = uuid4()
             self.author = author
+            if self.pinned:
```
jaywink: Hmm this whole `elif author:` is kinda weird, had to dig to find out why it isn't just an `else:`. Anyway, your code should run even if no author. So maybe restructure like:

    if self.pk:
        ...
    else:
        if self.pinned:
            ...
        if author:
            ....

socialhome/content/models.py
```diff
@@ -118,6 +118,8 @@ def save(self, author=None, *args, **kwargs):
...
+                self.order = Content.objects.filter(pinned=True).count() + 1
```
jaywink: Your query is wrong. We don't want the count of pinned content, we want the next unused value of `order`. So you need the maximum current value and then `+1`.

jaywink (changes requested):

socialhome/content/models.py
christophehenry: So `Content.objects.count() + 1`?

jaywink: No, that would be the count of objects. You need the largest value. Check https://docs.djangoproject.com/en/1.10/ref/models/querysets/#max

socialhome/users/tests/test_views.py
```diff
@@ -146,6 +148,18 @@ def test_contents_queryset_returns_content_in_correct_order(self, admin_client,
         assert qs[1].id == contents[1].id
         assert qs[2].id == contents[0].id
 
+    def test_content_saved_in_correct_order(self, admin_client, rf):
```
jaywink: The feature is in the model, so the test should be in content app `test_models`.

socialhome/users/tests/test_views.py
```diff
@@ -146,6 +148,18 @@ def test_contents_queryset_returns_content_in_correct_order(self, admin_client,
...
+        request, view, contents, profile = self._get_request_view_and_content(rf, create_content=False)
+        
+        pinned_content_1 = Content(pinned=True, text="foobar")
+        pinned_content_2 = Content(pinned=True, text="foobar")
+        pinned_content_3 = Content(pinned=True, text="foobar")
+        pinned_content_1.save(author=profile)
+        pinned_content_2.save(author=profile)
+        pinned_content_3.save(author=profile)
+
+        assert {pinned_content_1.order, pinned_content_2.order, pinned_content_3.order} == {1, 2, 3}
```
jaywink: A Python set is not ordered, so this guarantees only that the elements are in the set: https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset

jaywink (changes requested):

christophehenry:
The problem is that it now overrides the `order` parameter set at creation. Like [here](https://github.com/jaywink/socialhome/blob/master/socialhome/users/tests/test_views.py#L79). I don't know how to solve.

socialhome/content/models.py
```diff
@@ -114,10 +115,17 @@ def save(self, author=None, *args, **kwargs):
...
-            self.guid = uuid4()
-            self.author = author
+        else:
+            if author:
+                # New with author, set a GUID and author
+                self.guid = uuid4()
+                self.author = author
+
+            if self.pinned:
+                max_order = Content.objects.values_list("order").distinct().aggregate(order=Max("order"))["order"]
```
jaywink: Way shorter and a lot more performant is:

    Content.objects.aggregate(Max("order"))["order__max"]

But, you're taking *all* content here. We only care about the content of current user. And only pinned content.

Ps pro-tip: use the `python manage.py shell_plus` command to locally play around with these kind of things before writing code.

socialhome/users/tests/test_models.py
```diff
@@ -102,3 +103,14 @@ def test_get_last_name(self):
         assert profile.get_last_name() == ""
         profile.name = ""
         assert profile.get_last_name() == ""
+
+    def test_content_saved_in_correct_order(self, admin_client, rf):
```
jaywink: This is in `socialhome/users`. It should be in `socialhome/content` app `TestContentModel`.

socialhome/users/tests/test_views.py
```diff
@@ -2,6 +2,7 @@
 from django.contrib.auth.models import AnonymousUser
 from django.core.urlresolvers import reverse
 from django.test import RequestFactory
+from factory import fuzzy
```
jaywink: Unused import.

jaywink (changes requested):

jaywink:
> The problem is that it now overrides the order parameter set at creation. Like here. I don't know how to solve.

Right, good point. The old tests need to be fixed. Easiest is to after creating the content, update it with the correct order we want, like this:

    content = ContentFactory(author=profile, pinned=True)
    Content.objects.filter(id=content.id).update(order=1)

socialhome/content/models.py
christophehenry: > Way shorter and a lot more performant is:
> ```
> Content.objects.aggregate(Max("order"))["order__max"]
> ```

I believe you but I don't understand. I remember during my studies my teachers telling me that filtering content like using `DISTINCT` improved queries performance.

socialhome/users/tests/test_models.py
```diff
@@ -3,6 +3,7 @@
 
 from socialhome.users.tests.factories import ProfileFactory, UserFactory
 from socialhome.users.utils import get_pony_urls
+from socialhome.content.models import Content
```
jaywink: Unused import.

socialhome/content/models.py
```diff
@@ -114,10 +115,17 @@ def save(self, author=None, *args, **kwargs):
...
-            self.guid = uuid4()
-            self.author = author
+        else:
+            if author:
+                # New with author, set a GUID and author
+                self.guid = uuid4()
+                self.author = author
+
+            if self.pinned:
+                max_order = Content.objects.aggregate(Max("order"))["order__max"]
```
jaywink: OK this is closer, but it still needs to get content only for *user who created this current content*, not all content in the database. You need to add a filter before the aggregate to limit to current authoring user.

jaywink (changes requested):
One last

christophehenry:
Done.

jaywink (approved):
Great, thanks!

Issue #116, opened 2017-03-10
Don't show signup link if signups are closed


jaywink:
Hide the link from the navbar if signups are closed.

PR #119, opened 2017-03-13
Don't show signup link if signups are closed (#116)

christophehenry:


christophehenry:
I'm not sure I got everything right since I never set the `is_open_for_signup` variable, but I didn't find any base class view.

socialhome/templates/base.html
```diff
@@ -50,9 +50,11 @@
                               <a class="nav-link" href="{% url 'account_logout' %}">{% trans "Logout" %}</a>
                           </li>
                       {% else %}
+                          {% if is_open_for_signup %}
```
jaywink: Where did you find this - it works? If you turn signups on, the link will show? Can you write a test for it? As you can see from the coverage, the part inside this `if` is now not executed in tests, which means either our tests run with signups disabled, or this code doesn't work.

The latter could be because the template engine by default doesn't raise on undefined variables, which is kind of confusing, should probably change that.

christophehenry: No, I don't think it works. That's the thing: I don't know where to inject the variable.

socialhome/templates/account/signup.html
```diff
@@ -13,6 +13,7 @@
 
       <p>{% blocktrans %}Already have an account? Then please <a href="{{ login_url }}">sign in</a>.{% endblocktrans %}</p>
 
+      {% if is_open_for_signup %}
```
jaywink: No need to check this - if the signup is closed it will show "signups are closed" anyway.

selection_342

socialhome/templates/account/login.html
```diff
@@ -13,9 +13,12 @@
       <h2>{% trans "Sign In" %}</h2>
       {% get_providers as socialaccount_providers %}
       {% if socialaccount_providers %}
-      <p>{% blocktrans with site.name as site_name %}Please sign in with one
-      of your existing third party accounts. Or, <a href="{{ signup_url }}">sign up</a>
-      for a {{ site_name }} account and sign in below:{% endblocktrans %}</p>
+      <p>{% blocktrans %}Please sign in with one
+      of your existing third party accounts.{% endblocktrans %}
+        {% if is_open_for_signup %}
```
jaywink: Same here, no need to touch this file. Just need to make sure the nav bar is not rendered with the link.

jaywink (changes requested):

socialhome/templates/base.html
jaywink: Please test things locally before submitting a PR ;)

If you check how the signup template/view figures out that the user cannot sign up, that should point you in the right direction. There is a clear message that comes up when going to `/accounts/signup/` when signups are closed, so the view knows that from somewhere.

christophehenry: > Please test things locally before submitting a PR ;)

It's an habit I got from contributing to diaspora. I like to open early PRs everything I write is public and I can get some advice. But you're right: I should have made clear that it's WIP.

christophehenry:
\o/

socialhome/tests/test_views.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+
+@pytest.mark.usefixtures("db")
```
jaywink, line 4: No need for the db fixture as nothing is created to the database.

jaywink (changes requested):
Looking purrrrfect :) Except a small detail in the test. Quick change to that and good to merge!

socialhome/tests/test_views.py
christophehenry: The test was failiing with PyTest advising me to add it.

jaywink: Oh well, I could be wrong, that is fine then 👍 Thanks, merging!

jaywink (approved):

Issue #124, opened 2017-03-19
Don't overflow images in content create preview


jaywink:
Content create preview should have a `img: max-width: 100%;` to stop images overflowing the preview area.

imshashwataggarwal:
Hi @jaywink , Could you guide me how to get going on the issue.

jaywink:
Closed via #126 

Sorry @imshashwataggarwal - this got fixed already. Maybe you would be interested in some other issue?

PR #126, opened 2017-03-25
Fix image from overflowing in content create preview

winjar:
Fix image from overflowing in content create preview

Refs: #124


jaywink (approved):
Thanks! <3

PR #191, opened 2017-07-13
Adding Webpack

christophehenry:


jaywink:
I really have nothing to add to this, shall I merge it in already or is there something you wanted to add still? Great stuff, getting this ball I mean mountain rolling :P As talked IRL, need to update dev documentation on "why" to not confuse someone who might look at the project and ponder about contributing. I can document some thoughts and plans according to what we discussed in another PR 👍 

I don't think we need to add to production installation instructions yet until there is functionality behind the client app.

jaywink (approved):
Yay! 👍

christophehenry:
I have nothing to add. This setup is currently being tested on an incoming PR using VueJS. ES6 comilation using Babel works well. There will be probably configuration changes coming along with porting Grunt stuff to Webpack but I think this setup is good for the time being.

jaywink:
I added some design principles regarding the Vue.js work to http://socialhome.readthedocs.io/en/latest/roadmap.html - what do you think @christophehenry ? We can discuss these on IRC or here in issues. Finally had time to study things a bit and put down some thoughts from our hackathon :) Will fill in more details as we go ahead.

I'll anyway merge this as is, I'll provide feedback on the next PR you have open on things that I think could be changed.

Thanks for kicking this off! <3

christophehenry:
> Thanks for kicking this off! <3

You're very welcome ;)

> what do you think @christophehenry ?

Just a quick mention about search system that I think I already mentionned during the hackathon regarding ElasticSearch. It's a Java tool. We have to consider the possibility of user installing our application on simple hostings like [Gandi's](https://www.gandi.net/hosting/simple?language=python) with no acces whatsoever to admin console or installing on a small computer not powerful enough to run both a Python and a Java VM. So ElasticSearch binding should stay optional.

Another possibility is to provide a default fallback from ElasticSearch to a custom search engine based on [Lucene](https://en.wikipedia.org/wiki/Apache_Lucene) on which ElasticSearch is based and provides Python bindings. But this means a lot more work.

PR #192, opened 2017-07-13
Add search for profiles

### LINKED ISSUE #163: Add searching for profiles
When searching handles:
* If the profile is found locally, give that result.
* If not, do a remote lookup.

When searching text strings:
* Search in local profiles for matches with name

jaywink:
The search input in the navbar allows searching for profiles via partial match.

selection_371

TODO:
* [x] Tests
* [x] Implement remote fetch if a full handle wasn't found locally. By full handle, we mean something that validates to an email address
* [x] Make search input auto focus on search page
* [x] Add documentation to installation and dev for rebuild_indexes
* [x] Make ansible role rebuild indexes on first install
* [x] Add signals to update search index on profile saves
* [x] Should some profiles be filtered out? `SITE` for not logged in? `LIMITED` would make sense but tricky to do correctly, or should we change things to render those but show limited information instead, instead of complete hide? Log refactoring issue if so.
* [x] Make partial word matches work
* [x] Index only username part of handle, not pod name
* [x] On direct match go directly to profile

Closes #163


jaywink:
ping @christophehenry - maybe you want to check PR's, if not for nothing else then keeping up to date what I'm adding, if you plan on working on the frontend stuff?

PR #193, opened 2017-07-16
Partial front-end rewriting using VueJS

christophehenry:
This is a continuation of #191. It starts rewriting the front-end using VueJS, Vuex stores and VueResource.


jaywink:
OK, jumping on this now 👍 Sorry for the wait.

So, I've been thinking about things and tried to dump some of those thoughts in the [documentation](http://socialhome.readthedocs.io/en/latest/roadmap.html). Basically, I feel that instead of `socialhome/client` as the location for the streams Vue code, it should be based in the `streams` app. This is more logical and if we have other Vue apps, they can also live in their "own apps". If one day everything becomes a Vue app - then we just pull everything out to one place. Right now that looks unlikely and unnecessary.

The other thing is that you can't add anything to the base template at this stage. Otherwise we pretty much can't merge this in until it covers the whole stream use case, which could take a year.

Some ideas/todo items to move forward:
* [x] Move `socialhome/client` code under `socialhome/streams/app` as indicated by [docs](http://socialhome.readthedocs.io/en/latest/roadmap.html#code-layout).
* [x] Remove the current additions to the base and home templates.
* [x] Add a `socialhome/streams/templates/streams/app.html` template that extends the `base.html` like all templates, but then injects the Vue app into `content` block - just like you're doing in `home.html` now.
* [x] Load the new `webpack.bundle.js` only in `app.html` by placing it in the `javascript` block after a `{{ block.super }}` call ([docs](https://docs.djangoproject.com/en/1.10/ref/templates/language/#template-inheritance)).
* [x] Add to `socialhome.streams.views.BaseStreamView` code to catch if the request has an extra `?vue=1` or similar parameter. If so, dynamically change the template name that is rendered to `streams/app.html`. This way we can use the Vue version with the same context and all stream views by passing in an extra URL parameter.

Does this sound like a good plan? I had an idea that we could pass some initial serialized to JSON stream object data to the context and blurp it out in the template for Vue. This would do two things: 1) Make the stream load faster on initial load (= no need for extra fetch from backend) and 2) Allow prototyping the stream Vue code with actual data without the API endpoints. What do you think?

christophehenry:
Just a quick note regarding your [architecture](http://socialhome.readthedocs.io/en/latest/architecture.html#vue-js-app-s). Webpack still needs only one JS file as input that imports everything (see [here](https://github.com/jaywink/socialhome/blob/master/webpack.config.js#L5)), so we need to add a `main.js` file at the root of the `socialhome` directory:

```
socialhome/
  **main.js**
  streams/
    app/
      components/
        (components)
      App.vue
      main.js
    templates/
      streams/
        app.html
    views.py
```

jaywink:
Well, I was thinking that the streams would be a vue app itself. The entire frontend will not be (at least for a long time) rewritten in Vue. So I don't see why there should be a webpack bundle for the main socialhome app? If we were planning to rewrite the whole frontend with vue then yeah maybe, but I guess the structure would then look very different anyway. The point of placing the `main.js` in the streams app was that it would be a JS application doing *only* streams stuff - as such, it would live in the streams app.

christophehenry:
Isn't the plan to rewrite the whole front-end in Vue and only expose an HTTP API on the back-end?

jaywink:
Maybe in the very long term future, but initially I think it makes sense to rewrite the streams separately. Otherwise this rewrite will take years :P And starting from the other pages makes no sense since *they work fine*. The streams is where the problems and the need for rewrite is.

jaywink (changes requested):
Looking better! 👍 

Basically though, the old stream has to work until the whole streams rewrite is finished. So this and further PR's must keep the old stream as is and make changes only to the new stream.

In comment https://github.com/jaywink/socialhome/pull/193#issuecomment-317281378 I gave notes on how to do this. Basically the stream should have separate templates for the new stream. These templates would be loaded IF for example `?vue=1` is passed to the view.

Whether there should be a `app_followed.html` and `app_public.html` for example or just one `app.html` I'm not entirely sure. I guess the Vue code could contain those strings instead of the Django templates. All you really need to do then is to export the `stream_name` to the vue templates, and any data of course. But the texts would be in the Vue templates themselves.

Does that make sense?

socialhome/streams/app/components/StampedElement.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+    <div>
+        <h2>{{ stampedContent.h2 }}</h2>
+        <p>{{ stampedContent.p }}</p>
+    </div>
+</template>
+
+<script>
+    import Vue from "vue";
```
jaywink: Do you mind if we ditch the semicolons for more Pythonic javascript? I would very much like that.

Need to make an ESlint file btw.. I put some ideas to use Airbnb style without semicolons in [architecture notes](http://socialhome.readthedocs.io/en/latest/architecture.html#code-style).

socialhome/templates/base.html
```diff
@@ -102,6 +102,7 @@
         <script type="text/javascript" src="{% url 'javascript-catalog' %}"></script>
         <!-- Compiled project JS -->
         <script src="{% static 'js/project.js' %}"></script>
+        <script src="{% static 'js/webpack.bundle.js' %}"></script>
```
jaywink: Please add this as a super call in the template loading the streams app, like indicated in https://github.com/jaywink/socialhome/pull/193#issuecomment-317281378. That way it's only loaded on pages it needs to be.

socialhome/streams/templates/streams/base.html
```diff
@@ -5,11 +5,12 @@
 {% block containertype %}container-fluid{% endblock %}
 
 {% block ogtags %}
-    <meta property="og:title" content="{% block stream_title %}{% endblock %} - {{ request.site.name }}" />
-    <meta property="og:type" content="website" />
-    <meta property="og:url" content="https://{{ request.site.domain }}{{ request.path }}" />
+    <meta property="og:title" content="{% block stream_title %}{% endblock %} - {{ request.site.name }}"/>
```
jaywink: There are no actual changes related to this PR in this file, just reorganizing code. Can we leave it as is? Template code starts to look a bit ugly if you start breaking the lines too aggressively, and especially if it's multi-line but closing tags are not on the same level as the opening tag. In any case, since these changes are not related to the PR, it would be nicer to leave this file as it was.

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,41 @@
...
+{% load staticfiles %}
+
+{% block content %}
+    {# Temporary hack; to be removed #}
+    <script>
+        window.templateData = {
+            contentList: [
+                {% for content in content_list %}
+                    {
+                        content: "{{ content.rendered|safe }}",
```
jaywink: What if `content.rendered` has a `"` in it, won't this break the object in the JS side?

Since this data is quite complex, might be better to make a custom template tag to do a `json.dumps(obj)` as here: https://stackoverflow.com/a/25538871/1489738 .. Also `name` should be done the same.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+    <div>
+        <p v-html="post.content"></p>
+        <div class="grid-item-author-bar mt-1">
+            <div class="profile-box-trigger">
```
jaywink: Well, maybe not required in this PR, but this jQuery trigger class isn't really required here at all. We need Vue to watch for a click even on this div instead and then change some component state to make `profile-box` below visible. The same, `profile-box` class `hidden` should only render if this component local state defines that it should be hidden.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+    <div>
+        <p v-html="post.content"></p>
+        <div class="grid-item-author-bar mt-1">
```
jaywink: Actually, the whole author bar + profile box could be its own component 👍

socialhome/streams/app/stores/streamStore.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default new Vuex.Store({
+    state: {
+        translations: {
+            stampedContent: {
+                h2: window.templateData.stampedContent.h2,
+                p: window.templateData.stampedContent.p
+            }
+        },
+        posts: window.templateData.contentList
```
jaywink: We call this stuff `content`, not 'posts` ;)

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,29 @@
...
+            <div class="profile-box-trigger">
+                <img :src="post.author.imageUrlSmall" class="grid-item-author-bar-pic">
+            </div>
+            <div class="profile-box hidden">
+                {{ post.author.handle }}
+                <div class="pull-right">
+                    <a :href="post.author.absoluteUrl" class="btn btn-secondary"
```
jaywink: Multi-line tags look less confusing if the closing tag is on a brand new line like you've done in many other places. Also the `i` indentation should be one space more.

jaywink (commented):
A few tweaks to make. The splitting components and triggers are not necessarily todo for this PR, depends how much you want to get ready here. I'll pull in this and play around too. Would be nice to get some initial (very simple and basic!) tests in too, using the suggested Karma framework (https://vuejs.org/v2/guide/unit-testing.html). I would prefer Mocha over Jasmine - Karma supports both. Mocha is simple and we've already used it in this project. I don't mind playing around with this part too 👍

socialhome/streams/templates/streams/base.html
christophehenry: > There are no actual changes related to this PR in this file, just reorganizing code. Can we leave it as is?

The problem of having the reformat shortcut reflex :)
I fix that at once.

socialhome/streams/app/components/StreamElement.vue
christophehenry: > Vue to watch for a click even on this div instead and then change some component state to make `profile-box` below visible. 

I know. This PR is not finished at all. I first rewrite the template before taking care of the events ;)

christophehenry:
> I would prefer Mocha over Jasmine - Karma supports both. Mocha is simple and we've already used it in this project.

Mocha and Jasmine have the same syntax so I really don't mind ;)

christophehenry:
I think this has gone big enough. No need to produce a PR of 2000 lines that will be reviewed in 6 months. The `vue=true` parameter isolates the rewrite from the rest so there's no need for a complete work.

tmp_17140-img_20170806_151039-1502674340

I can has a review?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,74 @@
...
+            v-bind="author"
+            :is-user-author="isUserAuthor"
+            :is-user-authentificated="isUserAuthentificated"
+        ></author-bar>
+        <div class="grid-item-bar">
+            <div class="row">
+                <div class="col-9">
+                    <!-- TODO: Modal and Socket need a rewrite to get rid of grid-item-open-action -->
+                    <span
+                        v-if="edited"
```
jaywink: This `v-if` should be for the `{{ editedText }}` not for the whole timestamp. Now unless it is edited the whole timestamp is hidden.

socialhome/streams/app/main.js
```diff
@@ -0,0 +1,11 @@
+import Vue from "vue"
+import VueResource from "vue-resource"
+import "./stores/streamStore"
+import "./components/Stream.vue"
+
+Vue.use(VueResource)
+
+export default new Vue({
+    el: "#app",
+    http: {root: "/"}
```
jaywink: This doesn't seem right. Only one user set stream is at `/` - mostly streams are just everywhere. What is this for?

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,59 @@
...
+                    {
+                        htmlSafe: {{ content.rendered|json_safe }},
+                        author: {
+                            guid: "{{ content.author.guid }}",
+                            name: "{% if content.author.name %}{{ content.author.name }}{% else %}{{ content.author.handle }}{% endif %}",
+                            handle: "{{ content.author.handle }}",
+                            imageUrlSmall: "{{ content.author.safer_image_url_small }}",
+                            absoluteUrl: "{{ content.author.get_absolute_url }}",
+                            homeUrl: "{{ content.author.home_url }}",
+                            isUserFollowingAuthor: {% if not content.author.id in request.user.profile.following_ids %}true{% else %}false{% endif %}
```
jaywink: I think this should be without the `not`.

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,59 @@
...
+                        id: {{ content.id }},
+                        timestamp: "{{ formatted_timestamp }}",
+                        humanizedTimestamp: "{{ humanized_timestamp }}",
+                        edited: {{ content.edited|json_safe }},
+                        childrenCount: {{ content.children.count }},
+                        {% if content.author == request.user.profile %}
+                            isUserAuthor: true,
+                            updateUrl: "{% url " content:update" content.id %}",
+                            replyUrl: "{% url "content:reply" pk=content.id %}",
+                            deleteUrl: "{% url " content:delete" content.id %}",
```
jaywink: This should have additionally `?next={{ request.path }}` due to recent changes.

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,59 @@
...
+                            isUserFollowingAuthor: {% if not content.author.id in request.user.profile.following_ids %}true{% else %}false{% endif %}
+                        },
+                        id: {{ content.id }},
+                        timestamp: "{{ formatted_timestamp }}",
+                        humanizedTimestamp: "{{ humanized_timestamp }}",
+                        edited: {{ content.edited|json_safe }},
+                        childrenCount: {{ content.children.count }},
+                        {% if content.author == request.user.profile %}
+                            isUserAuthor: true,
+                            updateUrl: "{% url " content:update" content.id %}",
```
jaywink: Not required but better be explicit so `pk=content.id`.

socialhome/streams/templates/streams/vue.html
jaywink: Not required but better be explicit so `pk=content.id`.

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,59 @@
...
+                            deleteUrl: "",
+                        {% endif %}
+                    },
+                {% endfor -%}
+            ],
+            streamName: "{{ stream_name }}",
+            isUserAuthentificated: {% if request.user.is_authenticated %}true{% else %}false{% endif %},
+            showAuthorBar: {% if not stream_name|startswith:"profile_" %}true{% else %}false{% endif %},
+            {# Temporary hack; translation system to come #}
+            stampedContent: {
```
jaywink: To support other streams (now for example tags stream says "public" in the branch) you could move this object out into it's own for example `window.stampedContent`. And then in each of the Django templates that is rendered, set the correct texts.

The problem is some of the text is quite dynamic, for example profile stream. I wonder how that should be handled - it should really be vue components rendered `if streamName starts with "profile_"`. Maybe do that in follow up PR's and just leave this as is for now.

webpack.config.js
```diff
@@ -1,25 +1,44 @@
...
-     stats: {
-         colors: true
-     },
-     devtool: 'source-map'
- };
+module.exports = {
+    entry: path.resolve(__dirname, "socialhome/streams/app/main.js"),
+    output: {
+        path: path.resolve(__dirname, "socialhome/static/js"),
+        filename: "webpack.bundle.js"
```
jaywink: Can we call the file `streams.bundle.js` instead? I feel that would make more sense.

jaywink (changes requested):
Awesome work - the code looks so pretty compared to the jQuery mess currently <3 Even if it is JS!

Some things to fix, otherwise they will be hidden when merged so better fix these now. I will aim to list missing stuff in issue #202 as tick boxes hopefully tomorrow.

webpack.config.js
```diff
@@ -2,19 +2,20 @@ const path = require("path")
...
 
 module.exports = {
-    entry: path.resolve(__dirname, "socialhome/streams/app/main.js"),
+    entry: {
+        stream: path.resolve(__dirname, "socialhome/streams/app/main.js")
+    },
     output: {
         path: path.resolve(__dirname, "socialhome/static/js"),
-        filename: "webpack.bundle.js"
+        filename: "webpack.[name].js"
```
jaywink: Oh ok this is fine too 👍

karma.conf.js
```diff
@@ -0,0 +1,36 @@
+const wpConf = require("./webpack.config")
+wpConf.devtool = "inline-source-map"
+
+module.exports = config => {
+    let findTests = () => {
+        // THIS is why I hate NodeJS. That tool is so immature that you always
```
jaywink: Maybe remove the personal statement especially as it doesn't have a name ;)

socialhome/streams/app/tests/main.tests.js
```diff
@@ -0,0 +1,8 @@
+import {should} from "chai"
+should()
+
+describe("Test", () => {
+    it("is actually a test (RLY!?)", () => {
+        true.should.equal(true)
```
jaywink: Yay! 🎉 🍻

socialhome/streams/app/tests/main.tests.js
```diff
@@ -0,0 +1,8 @@
+import {should} from "chai"
+should()
```
jaywink: Why are we calling `should`?

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+    <div class="grid-item-author-bar mt-1">
+        <div @click.stop.prevent="profileBoxTrigger" class="profile-box-trigg">
```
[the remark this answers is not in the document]
christophehenry: Because `profile-box-trigger` is used on the jQuery part. Should I change the `profile-box-trigger` class to `jq-profile-box-trigger`?

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -0,0 +1,86 @@
...
+</template>
+
+<script>
+    import Vue from "vue"
+
+    export default Vue.component("author-bar", {
+        props: {
+            handle: {type: String, required: true},
+            name: {type: String, required: true},
+            guid: {type: String, required: true},
```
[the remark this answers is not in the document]
christophehenry: This is a style I got when I was writing Groovy: Spaces inside closures, no space inside dict (which, in Groovy writes `[attr: "val"]`).

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -0,0 +1,86 @@
...
+            profileBoxTrigger() {
+                this.showProfileBox = !this.showProfileBox
+            }
+        }
+    })
+</script>
+
+<style scoped lang="scss">
+    .profile-box-trigg {
+        cursor: pointer
```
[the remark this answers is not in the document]
christophehenry: Nope, it's SCSS actually. I think at some point, PyCharm complained about optionnal semi-colons, so I removed them.

socialhome/streams/app/tests/main.tests.js
christophehenry: Got from a quick read of  [the documentation](http://chaijs.com/guide/styles/#should). Did I get this wrong?

karma.conf.js
christophehenry: Can I keep it if I put my name?

socialhome/streams/app/main.js
christophehenry: It configures a library to make HTTP calls. It's not used yest, I can remove.

socialhome/streams/app/components/AuthorBar.vue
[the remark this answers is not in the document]
jaywink: Hmm, I can't seem to find a reference about SCSS not requiring semicolons. Do you have any? Maybe pycharm is just confused that it is SASS?

socialhome/streams/app/tests/main.tests.js
jaywink: Apparently not, though it looks freaking weird :P I guess do as they say then!

karma.conf.js
jaywink: Hehe. I'd rather not, might make some JS fan sad ;)

socialhome/streams/app/main.js
jaywink: I guess it's fine if the library is already included? I'll work on the streams API next week 👍

socialhome/streams/app/components/AuthorBar.vue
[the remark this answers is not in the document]
jaywink: True, nasty conflicts.. We could separate `streams.js` out from the compiled `project.js` and not load that here but hmm... Maybe call this version `profilebox-trigger` and leave the jQuery one alone?

karma.conf.js
christophehenry: K. Removing, then :)

socialhome/streams/app/tests/main.tests.js
christophehenry: It looks we can write `import "chai/should"`. Have to test.

socialhome/streams/app/components/AuthorBar.vue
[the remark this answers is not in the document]
christophehenry: >  I can't seem to find a reference about SCSS not requiring semicolons. 

Me neither. It ticked me at the time, but I didn't dig any further. Will add semi-colons.

karma.conf.js
```diff
@@ -0,0 +1,36 @@
...
+        // have to install a bunch of crappy modules to do some tiny operation
+        // Long life to Python \o/
+        let glob = require("glob")
+        return glob.sync("./socialhome/**/app/tests/**/*\.tests\.js")
```
christophehenry: I probably should have better read [the documentation](https://karma-runner.github.io/0.13/config/configuration-file.html):

>All of the configuration options, which specify file paths, use the minimatch library to facilitate flexible but concise file expressions so you can easily list all of the files you want to include and exclude.

This function seems unneeded.

karma.conf.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+        return result
+    }
+
+    config.set({
+        browsers:      ["PhantomJS"],
+        files:         findTests(),
+        frameworks:    ["mocha"],
+        plugins:       [
+            "karma-phantomjs-launcher",
```
christophehenry: According to [documentation](https://karma-runner.github.io/0.13/config/configuration-file.html), this is unneeded too:

> PhantomJS (launcher comes installed with Karma)

karma.conf.js
```diff
@@ -0,0 +1,36 @@
...
+            "karma-mocha",
+            "karma-sourcemap-loader",
+            "karma-webpack"
+        ],
+        preprocessors: preprocessors(),
```
christophehenry: Possibly unneeded.

jaywink: I love the smell of burning unnecessary code/config lines in the morning!

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,62 @@
...
+                            homeUrl: "{{ content.author.home_url }}",
+                            isUserFollowingAuthor: {% if not content.author.id in request.user.profile.following_ids %}true{% else %}false{% endif %}
+                        },
+                        id: {{ content.id }},
+                        timestamp: "{{ content.formatted_timestamp }}",
+                        humanizedTimestamp: "{{ content.humanized_timestamp }}",
+                        edited: {{ content.edited|json_safe }},
+                        childrenCount: {{ content.children.count }},
+                        isUserLocal: {% if content.author.user %}true{% else %}false{% endif %},
+                        contentUrl: "{{ content.get_absolute_url }}",
```
jaywink: Have you rebased your branch on top of master recently? `Content.get_absolute_url` method was only added some days ago, around the time I suggested you to use it ;)

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,62 @@
...
+                        {% if content.author == request.user.profile %}
+                            isUserAuthor: true,
+                            updateUrl: "{% url " content:update" content.id %}",
```
jaywink: Extra space between in `" content`

socialhome/streams/templates/streams/vue.html
jaywink: Extra space between in `" content`

socialhome/streams/app/tests/main.tests.js
christophehenry: Documentation is not up to date. But `import "chai/register-should"` works.

karma.conf.js
christophehenry: Documentation is shit. Does not work without it.

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+    return _slurp_json_context(context, lambda obj: str(obj))
+
+
+@register.simple_tag(takes_context=True)
+def slurp_json_context_unsafe(context):
+    """
+    Has the same behaviour than ``slurp_json_context`` at the exception
+    of throwing an error when a variable can't be serialized.
+    """
+    return _slurp_json_context(context, None)
```
jaywink: Instead of passing in `None` it's better to make the called function as follows: `_slurp_json_context(context, json_default=None)`.

This way, don't pass `json_default` if you want the `None` behaviour.

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+
+register = Library()
+
+
+@register.simple_tag(takes_context=True)
+def slurp_json_context(context):
+    """
+    Serializes ``context["json_context"]`` into JSON and generates a JS
+    script to attach it to ``window.context``
+    Args:
```
jaywink: Docstrings args etc should preferably be `reST` style, see from here how: https://stackoverflow.com/a/24385103/1489738

Also, Python convention is to structure docstrings like this:

```
"""This is one sentence describing the method.

This is a longer description which can span many lines and has many sentences or code.
There should be an empty line before the main body part. The closing """ is on a new line,
unless the whole docstring is just one line, then on the same line. An empty line before :param: etc.

:param foo: Nice param
"""
```

See also PEP-257: https://www.python.org/dev/peps/pep-0257/

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
+from django.utils.safestring import mark_safe
+from django.template import Library
+import json
+
+register = Library()
+
+
+@register.simple_tag(takes_context=True)
+def slurp_json_context(context):
```
jaywink: I'm not a huge fan of the method name. Maybe just `json_context`?

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+        context: Current view context
+
+    Returns: "<script>window.context = {<context["json_context"]>}</script>"
+
+    """
+    return _slurp_json_context(context, lambda obj: str(obj))
+
+
+@register.simple_tag(takes_context=True)
+def slurp_json_context_unsafe(context):
```
jaywink: This isn't actually used. Adding code that *might* be used in the future leads to dead code which never gets removed. Let's add this when we need it :)

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Has the same behaviour than ``slurp_json_context`` at the exception
+    of throwing an error when a variable can't be serialized.
+    """
+    return _slurp_json_context(context, None)
+
+
+def _slurp_json_context(context, json_default):
+    if context.get("json_context", None) is None:
```
jaywink: `.get()` default is `None` or check against `None` since empty is falsey, so you don't need to specify it, just do `if context.get("json_context"):`.

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+        return ""
+
+    json_dump = json.dumps(context["json_context"], default=json_default)
+    return mark_safe("<script>window.context = %s</script>" % json_dump)
```
jaywink: `%s` should probably be followed by a `;` since this is browser JS.

socialhome/streams/templatetags/slurp_json_context.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Serializes ``context["json_context"]`` into JSON and generates a JS
+    script to attach it to ``window.context``
+    Args:
+        context: Current view context
+
+    Returns: "<script>window.context = {<context["json_context"]>}</script>"
+
+    """
+    return _slurp_json_context(context, lambda obj: str(obj))
```
jaywink: Why do we need the `str(obj)`?

socialhome/streams/views.py
```diff
@@ -10,12 +12,67 @@ class BaseStreamView(ListView):
     ordering = "-created"
     paginate_by = 30
     stream_name = ""
+    vue = False
+    queryset = []
```
jaywink: Why has this been added? It's a super bad idea to initialize lists as class attributes. Also, `queryset` is not going to contain a list, but a `QuerySet`.

But I don't see why it was necessary to add this?

socialhome/streams/views.py
```diff
@@ -10,12 +12,67 @@ class BaseStreamView(ListView):
...
+        return super().dispatch(request, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["stream_name"] = self.stream_name
+        if self.vue:
+            context["json_context"] = self._define_context_json(self.stream_name)
         return context
 
+    def _define_context_json(self, stream_name):
```
jaywink: Please add a  `# pragma: no cover` since this is temporary code and we are not going to have tests for it and so coverage shouldn't be affected. See https://coverage.readthedocs.io/en/coverage-4.4.1/excluding.html

socialhome/streams/views.py
```diff
@@ -10,12 +12,67 @@ class BaseStreamView(ListView):
...
+        request_user_profile = getattr(self.request.user, "profile", None)
+
+        def dump_content(content):
+            is_user_author = (request_user_profile is not None and content.author == request_user_profile)
```
jaywink: Just `content.author == request_user_profile` should be enough, because if the latter is `None` - they're not equal.

socialhome/streams/views.py
```diff
@@ -10,12 +12,67 @@ class BaseStreamView(ListView):
...
+                "contentUrl": content.get_absolute_url(),
+                "isUserAuthor": is_user_author,
+                "updateUrl": reverse("content:update", kwargs={"pk": content.id}) if is_user_author else "",
+                "replyUrl": reverse("content:reply", kwargs={"pk": content.id}) if is_user_author else "",
+                "deleteUrl": reverse("content:reply", kwargs={"pk": content.id}) if is_user_author else ""
+            }
+
+        return {
+            "contentList": [dump_content(content) for content in self.queryset],
+            "currentBrowsingUserId": getattr(request_user_profile, "id", ""),
```
jaywink: As said, `currentBrowsingProfileId` would be more exact.

socialhome/streams/views.py
```diff
@@ -10,12 +12,67 @@ class BaseStreamView(ListView):
...
+            "streamName": stream_name,
+            "isUserAuthentificated": bool(self.request.user.is_authenticated),
+            "showAuthorBar": stream_name.startswith("profile_"),
```
jaywink: This should be `not stream_name.startswith("profile_")`? Unless I'm reading it wrong.

Basically author bar should be shown "if the stream is not a profile stream".

socialhome/streams/templates/streams/vue.html
```diff
@@ -0,0 +1,15 @@
+{% extends "streams/base.html" %}
+{% load staticfiles slurp_json_context string_utils %}
+
+{% block content %}
+    {% slurp_json_context %}
```
jaywink: I think this should be after the `{{ block.super }}` in the javascripts section, before the `webpack` bundle load, right? Otherwise it is slowing the page load it the browser has to process script tags here. Minor but no reason not to have it later.

socialhome/streams/templatetags/slurp_json_context.py
jaywink: This template tag should have a few very small basic tests if it is permanent code.

Tests should be placed by (our) convention in `streams/tests/test_templatetags.py`. See test notes here: http://socialhome.readthedocs.io/en/latest/contributing.html#tests .. just a few simple tests to ensure basic behaviour - you don't need to create more complex data.

jaywink (changes requested):
A few more comments. Looking good, hoping to merge soon! ;)

socialhome/streams/templatetags/slurp_json_context.py
christophehenry: Yet another markup language to learn :'(

socialhome/streams/templatetags/slurp_json_context.py
christophehenry: For classes that can't be implicitely serialized into JSON. I had the case for `LazyBooleanValue` or something like that. I don't remember exactly which variable it is.

socialhome/streams/templatetags/slurp_json_context.py
christophehenry: No, it is. I used it for debugging puposes. When I ported the code from the template I forgot to add parenthesis to seme function calls and had a nasty string representation in my JSON code because of `lambda obj: str(obj)`.

socialhome/streams/templatetags/slurp_json_context.py
christophehenry: True. Some reflexes I don't have yet :)

socialhome/streams/templatetags/slurp_json_context.py
christophehenry: Most likely, yes. Nice catch, thx.

socialhome/streams/views.py
christophehenry: I don't remember. It must be an artefact of my fail/retry coding proces. Sry :(

socialhome/streams/views.py
christophehenry: It's not temporary code :)

christophehenry: BTW, maybe I just should move that code into `socialhome.templatetags`.

package.json
```diff
@@ -22,13 +24,32 @@
...
+    "vue-loader": "^12.2.2",
+    "vue-style-loader": "^3.0.1",
+    "vue-template-compiler": "^2.4.1",
     "webpack": "^3.2.0"
   },
   "engines": {
     "node": ">=0.8.0"
+  },
+  "dependencies": {
+    "bootstrap": "^3.3.7",
```
jaywink: Hmmm this was probably added automatically? We use bootstrap 4 - which is already installed by bower. If it needs to be installed here too (probably), it should be the 4 version?

socialhome/streams/app/components/Stream.vue
```diff
@@ -0,0 +1,39 @@
...
+            <div class="gutter-sizer"></div>
+            <stream-element
+                class="grid-item"
+                v-for="content in $store.state.contentList"
+                v-bind="content"
+                :key="content.id"
+                :data-content-id="content.id"
+                :is-user-authentificated="$store.state.isUserAuthentificated"
+                :show-author-bar="$store.state.showAuthorBar"
+                :current-browsing-user-id="$store.state.currentBrowsingUserId"
```
jaywink: The `author-bar` component lists `currentBrowsingProfileId: {type: String, required: true},` but here I think you've not changed it yet.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+    <div>
+        <div v-html="htmlSafe"></div>
+        <author-bar
+            v-if="showAuthorBar"
+            v-bind="author"
+            :is-user-author="isUserAuthor"
+            :is-user-local="isUserLocal"
+            :current-browsing-user-id="currentBrowsingUserId"
```
jaywink: Also here variable rename needed.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -0,0 +1,75 @@
...
+            updateUrl: {type: String, required: true},
+            deleteUrl: {type: String, required: true},
+            replyUrl: {type: String, required: true},
+            childrenCount: {type: Number, required: true},
+            edited: {type: Boolean, required: true},
+            isUserLocal: {type: Boolean, required: true},
+            isUserAuthor: {type: Boolean, required: true},
+            showAuthorBar: {type: Boolean, required: true},
+            isUserAuthentificated: {type: Boolean, required: true},
+            currentBrowsingUserId: {type: String, default: ""}
```
jaywink: Here too.

socialhome/streams/templatetags/json_context.py
```diff
@@ -0,0 +1,39 @@
...
+    script to attach it to ``window.context``
+
+    :param context: Current view context
+    :return: ``<script>window.context = {<context["json_context"]>}</script>``
+    """
+    return _json_context(context, lambda obj: str(obj))
+
+
+@register.simple_tag(takes_context=True)
+def json_context_unsafe(context):
```
jaywink: I really would like this to be removed if it's only for debugging purposes. It is introducing code that is never executed into the project. If you really want, you could add a `raise_error=False` parameter to `json_context` function and use that to decide whether to raise. This is quite a common way to control functions that can either silently "do a bit more" or be strict and fail early. As long as you then write a test with and without the raise parameter.

socialhome/streams/templatetags/json_context.py
```diff
@@ -0,0 +1,39 @@
...
+    of throwing an error when a variable can't be serialized.
+
+    This function must be used for debug purposes to find variables
+    that can't implicitely serialized in JSON.
+    """
+    return _json_context(context)
+
+
+def _json_context(context, json_default=None):
+    if context.get("json_context") is None:
```
jaywink: `if not context.get("json_context"):` is more pythonic :)

jaywink (changes requested):
Only a few comments left, then it's mergy mergy time! 🎉

socialhome/streams/views.py
jaywink: > It's not temporary code :)

Well, it will be gone when the streams API comes? ;)

jaywink (approved):
I can't find anything else to add to this. Tests would be nice as discussed, but I think the whole WIP nature of this allows us to merge in and write tests while the Vue streams app stabilizes and the API is ready.

Good work, nice to see this going forward. Thanks! <3

christophehenry:
Oh <3

PR #231, opened 2017-08-06
Vue modal

christophehenry:
Adding a modal system on top of #193. Will be useful in a near future. There may be a simpler method, but I'm still learning Vue. Will refactor if I find it can be simplified later.

- [x] My awesome modal
- [ ] I can has tests?

An example of usage:

```html
<!-- This is how you define your modal template -->
<template>
    <modal>
        <!-- The element with attribute `slot="title"` will be your title -->
        <div slot="title">
            <h1>This is a modal</h1>
            <p>With a complex title</p>
            <p>If there's no title element, only the close button is rendred</p>
        </div>

        <!--
            Anything that comes inside your modal with not `slot="title"` 
            or `slot="footer"` attribute will be your content
        -->
        <div>
            <p>Here lies the body content of the modal</p>
        </div>

        <!--
             And, obviously, the element with attribute `slot="footer"`
             will be your footer
        -->
        <div slot="footer">
            <p>There even may be a footer...</p>
            <p>But may be not, up to you</p>
        </div>
    </modal>
</template>

<!-- 
    You can add whatever methods you please.
    If you define a modal with a complex form inside, 
    for instance, your logic comes here
-->
<script>
    import Vue from "vue"
    import "../../../shared/components/Modal.vue"

    export default Vue.component("test-modal", {})
</script>
```

Then, in the component that wants to show the modal:

```js
import {modalView} from "../main"
import TestModal from "./TestModal.vue"

onClickBtn(){
    // Here's how you display the modal
    modalView.renderModal(TestModal)
}
````

And the result:

screen shot 2017-08-06 at 10 51 32


christophehenry:
Stopping to work on this for now. Will dig it up when I actually need modals.

jaywink:
I'll close this PR here, otherwise it feels like ongoing work. Don't delete the branch ;)

PR #236, opened 2017-08-08
Allow user to change profile picture.

### LINKED ISSUE #151: Allow user to set profile picture
Allow uploading a picture as profile picture.

Additionally, support Gravatar pictues?

jaywink:
Profile menu now has an extra option "Change picture". This allows uploading a new picture and optionally setting focus point for cropping a picture that is not square shape.

* [x] Add tests

Closes #151


socialhome/utils.py
```diff
@@ -0,0 +1,7 @@
+from django.conf import settings
+
+
+def get_full_media_url(path):
+    return "{url}{media}{path}".format(
+        url=settings.SOCIALHOME_URL, media=settings.MEDIA_URL, path=path,
```
christophehenry, line 6: Maybe we could use f-strings?

jaywink: Would love to but need to support Python 3.4 and 3.5 for a while still. Ubuntu, which is easily one of the most popular installation servers, ships 3.4 in 14.04 which is still supported.

christophehenry: Meh. So many cool features fading away :'(

PR #254, opened 2017-08-13
Partial front-end rewriting using VueJS vol. 2

christophehenry:


socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -80,7 +80,8 @@
             profileBoxTrigger() {
                 this.showProfileBox = !this.showProfileBox
             }
-        }
+        },
+        updated() { Vue.redrawVueMasonry() }
```
jaywink: Would prefer a slightly more verbose but clearer syntax (which Airbnb style also mandates), so:

    updated() { 
        Vue.redrawVueMasonry()
    },

socialhome/streams/app/components/Stream.vue
```diff
@@ -35,5 +34,20 @@
...
+        data() {
+            return {
+                masonryOptions: {
+                    "item-selector":       ".grid-item",
+                    "column-width":        ".grid-sizer",
+                    "gutter":              ".gutter-sizer",
+                    "percent-position":    true,
+                    "stamp":               ".stamped",
+                    "transition-duration": "0s",
+                    "stagger":             0
```
jaywink: We don't have an ESlint config yet, but soon will once I add one - and dangling commas will be there since they make diffs better and look pythonic. So good idea to start getting used to adding commas on the last multi-line lines! :)

jaywink (changes requested):
A few very minor comments 👍 

I wonder if this plugin supports the [images loaded](https://imagesloaded.desandro.com/) script hmm... It's kind of mandatory, we will need to integrate it somehow, since images completing loading cause the grid to change size. But doesn't need to be in this PR 👍

christophehenry:
> I wonder if this plugin supports the images loaded script hmm...

What's awesome with VueJS is that [there's always a plugin](https://github.com/David-Desmaisons/Vue.ImagesLoaded).

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -46,41 +40,66 @@
 <script>
     import Vue from "vue"
 
+
     export default Vue.component("author-bar", {
-        props: {
-            handle: {type: String, required: true},
-            name: {type: String, required: true},
-            guid: {type: String, required: true},
+        props:    {
```
jaywink: Hmm please no extra spaces to align things - it makes diffs annoying when the longest element changes and it looks weird.

Need to add eslint and flake to travis so that builds fail on code style issues - better for contributors also to get instant feedback 👍

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -46,41 +40,66 @@
...
+            imageUrlSmall:            {type: String, required: true},
+            isUserAuthor:             {type: Boolean, required: true},
+            isUserLocal:              {type: Boolean, required: true},
+            isUserFollowingAuthor:    {type: Boolean, required: true},
+            isUserAuthentificated:    {type: Boolean, required: true},
         },
         data() {
-            return {showProfileBox: false}
+            return {
+                showProfileBox:         false,
```
jaywink: Same here, too many spaces.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -46,41 +40,66 @@
...
+                localIsUserFollowingAuthor: this.isUserFollowingAuthor,
```
jaywink: Vue doesn't distinguish state and props? That is a bit sad :(

How about `_isUserFollowingAuthor` so prefixing local state with `_`? Prefixing with `local` sounds a bit dirty, since a user can be "local" etc. The word has strong other meanings. Might end up with a variable `localLocal...` somewhere.

jaywink (changes requested):
Looks awesome, a few comments!

I think it's time to start adding tests to the component `methods` at minimum 👍

socialhome/streams/app/components/AuthorBar.vue
christophehenry: > How about _`isUserFollowingAuthor` so prefixing local state with `_`?

Variables prefixed with an underscore have a particular behaviour in VueJS. I discovered that before using the `local` prefix. They won't update you view when they change unless you explicitely use `this.$data._isUserFollowingAuthor`.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -44,48 +40,74 @@
 </template>
 
 <script>
-import Vue from "vue"
```
christophehenry: @jaywink: Please keep `<script>` content indented. `<template>` one is. It feels weird that `<script>` is not.

.babelrc
```diff
@@ -0,0 +1 @@
+{"presets": ["env"]}
```
christophehenry: This is weird. I just spent 2 hours trying to figure why my Webpack wasn't correctly transorming one the the peer dependancies during the test (vue-masony that uses the `const` keyword which is not understood by PhantomJS). The only way to solve the problem was by adding this file. This should not happen as the Babel configuration is already present in Webpack. I opened an issue upstream.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -61,31 +58,59 @@ export default Vue.component("author-bar", {
...
     },
     computed: {
+        isUserRemote(){
+            return !this.isUserLocal
+        },
         showFollowBtn() {
-            return this.isUserAuthentificated
-                    && !this.isUserLocal
-                    && !this.isUserFollowingAuthor
+            return this.canFollow && !this.$data._isUserFollowingAuthor
```
jaywink, line 73: Is `this.$data` the correct way to do this? Tried looking and can't find much. So trusting you on this 👍 Just looks weird and wondering if it should be used directly, since isn't it calculated by `data()` ?

socialhome/streams/app/stores/streamStore.js
```diff
@@ -1,22 +1,32 @@
...
             },
-        },
-        contentList: window.context.contentList,
-        streamName: window.context.streamName,
-        isUserAuthentificated: window.context.isUserAuthentificated,
-        showAuthorBar: window.context.showAuthorBar,
-    },
-})
+        }
+    }
```
jaywink: `catch` should be on the same line as previous `}`.

socialhome/streams/app/tests/main.tests.js
```diff
@@ -1,7 +1,13 @@
 import "chai/register-should"
+import "streams/app/main"
+
+import Vue from "vue"
+
 
 describe("Test", () => {
```
jaywink: This describe should probably be called something else. Maybe `Main`? Since we're testing what `main.js` does.

socialhome/streams/views.py
```diff
@@ -56,17 +56,18 @@ def dump_content(content):
...
-                    "p": gettext("Contains public content from around the network.")
-                }
-            }
+            "translations": {"stampedContent": self._get_stamped_content_translations()}
+        }
+
+    def _get_stamped_content_translations(self):
+        return {
+            "h2": gettext("Public"),
+            "p": gettext("Contains public content from around the network.")
```
jaywink: This way is ok for this PR but don't expand too much on it. We have to provide this content as a Vue component. The profile streams are way too complex to be done like this, they have user actions and stuff.

I think (after this PR) the stream stamped area should be a Vue component, one each for each stream type, and then we just decide which one to render depending on the stream name.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -61,31 +58,59 @@ export default Vue.component("author-bar", {
...
     },
     data() {
-        return {showProfileBox: false}
+        return {
+            showProfileBox: false,
+            _isUserFollowingAuthor: this.isUserFollowingAuthor,
+        }
     },
     computed: {
+        isUserRemote(){
```
jaywink: I think it would be good to add tests for the computed `isUserRemote`, `showFollowBtn`, `showUnfollowBtn`, `canFollow` and methods `profileBoxTrigger`, `unfollow`, `follow`. They don't look like they need much changing any more as they're behaving as they should 👍

jaywink (changes requested):

socialhome/streams/app/components/AuthorBar.vue
christophehenry: > Is `this.$data` the correct way to do this?

Yes. As told, earlier, the `_` prefixed data items work differently. A bit like the dunder methods in Python. They're not directly accessible. This is a convention to mark internal variable but it is enforced by the framework itself. It seems not documented, but this explenation is available [here](https://forum.vuejs.org/t/underscore-prefixes-for-vue-data/4018).

Though, I'm not confortable with this piec of code. It's basically a hack because properties must not be mutated. Props mutations get erased after each parent update. So passing this information via props seems not the way to go. A store, possibly? Or [`provide/inject`](https://vuejs.org/v2/api/#provide-inject)?

jaywink: Hmm. We can go with this, or, what about storing local state in `this.localState = {}` or something that isn't likely to conflict  with "props"?

socialhome/streams/views.py
christophehenry: I agree. But we need to find a way to make Django translation work.

socialhome/streams/app/components/AuthorBar.vue
christophehenry: > Hmm. We can go with this, or, what about storing local state in `this.localState = {}` or something that isn't likely to conflict with "props"?

It won't work either. Vue won't bind watcher deep into objects:

> **The object must be plain**: native objects such as browser API objects and prototype properties are ignored. A rule of thumb is that data should just be data - it is not recommended to observe objects with its own stateful behavior.

We would have to bind watchers manually.

> Yes. As told, earlier, the `_` prefixed data items work differently.

Actually, [it's written in the documentation](https://vuejs.org/v2/api/#data):

> Properties that start with `_` or `$` will not be proxied on the Vue instance because they may conflict with Vue’s internal properties and API methods. You will have to access them as `vm.$data._property`.

socialhome/streams/app/components/AuthorBar.vue
christophehenry: Done for `isUserRemote`, `showFollowBtn`, `showUnfollowBtn` and `canFollow`

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -61,31 +58,60 @@ export default Vue.component("author-bar", {
...
         },
         showUnfollowBtn() {
+            return this.canFollow && this.$data._isUserFollowingAuthor
+        },
+        canFollow() {
             return this.isUserAuthentificated
-                    && !this.isUserLocal
-                    && this.isUserFollowingAuthor
+                && !this.isUserAuthor
+                && this.isUserLocal
```
jaywink: Wait hmm why do we check if the user is local? Who is `isUserLocal`, the author or the logged in user? Anyway, remote users can be followed too.

socialhome/streams/app/tests/components/AuthorBar.tests.js
```diff
@@ -0,0 +1,81 @@
...
+                let target = new AuthorBar({propsData})
+                target.isUserRemote.should.be.false
+                propsData.isUserLocal = false
+                target = new AuthorBar({propsData})
+                target.isUserRemote.should.be.true
+            })
+        })
+
+        describe("canFollow", () => {
+            it("sould be true if user is local, authentificated and not the author", () => {
```
jaywink: So yeah, I think this (and the code) should be `should be true if user is authentificated and not the author` 👍

jaywink (changes requested):
One issue with the follow buttons and one comment not resolved (catch on wrong line) - I think then it's mergy time!

socialhome/streams/app/tests/components/AuthorBar.tests.js
```diff
@@ -0,0 +1,81 @@
...
+
+
+Vue.use(BootstrapVue)
+
+describe("AuthorBar", () => {
+
+    describe("computed", () => {
+        describe("isUserRemote", () => {
+            it("should be the opposite of property isUserLocal", () => {
+                let propsData = {
```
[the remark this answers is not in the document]
christophehenry: Yes, we can do that. I was also thinking about enabling JSX support in code and tests.

christophehenry:
> I think then it's mergy time!

This actually may never happen. Karma is not working anymore here. I'm tired trying to make JS crap to work, try to figure the correct setup and all. I spent 3 hours trying to make Karma working with Webpack yesterday. I solved the problem in a way that should not work. I report the problem upstream and I get shut down. And tonight, the whole crap do not work anymore even if I didn't modify any file.

Sometimes, my life feels so pointless and I feel really really empty... :'(

I need a hug...

christophehenry:
I just added Karma tests to Travis. Let's see what it says about it...

christophehenry:
Ok, so [yet another error](https://travis-ci.org/jaywink/socialhome/jobs/265473858#L1493). Communication between Karma and Webpack is complete shit. I'm out of I'm out of ideas. @jaywink, if you feel like trying to resolve the problem, a new eye may spot something I missed.

christophehenry:
That's pretty everything. I'm out of ideas. Travis seems to be ignoring my Karma configuration file, which specify that [Webpack should shut up](https://github.com/jaywink/socialhome/pull/254/files#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfbR28) but still lets [Webpack spits its logs](https://travis-ci.org/jaywink/socialhome/jobs/265493734#L1387).

mocha/fixtures/streams/AuthorBar.fixtures.js
```diff
@@ -0,0 +1,19 @@
+import _ from "lodash/core"
+
+const getPropsData = function (args = {}) {
```
jaywink: I kind of feel the fixtures should be located where the components and tests are, so in `socialhome/streams/app/tests/fixtures/`. Now they're isolated from both the code and the tests, which seems weird.

socialhome/streams/app/tests/components/AuthorBar.tests.js
```diff
@@ -0,0 +1,210 @@
...
+                    sinon.spy(console, "error")
+                    console.error.calledWith("Not logged in").should.be.true
+                    console.error.restore()
+                }))
+                target.instance().follow()
+            })
+        })
+    })
+
+    describe("lifcycle", () => {
```
jaywink: Typo, should be `lifecycle`? Also to be honest, the extra `describe` here is probably unnecessary, since it has another `describe` child.

jaywink (changes requested):
A few comments, then good for merging! Just push new commits, I'll squash them on my side before merging 👍

socialhome/streams/app/tests/components/AuthorBar.tests.js
christophehenry: Though I'd like to keep the `lifecycle`. I think it's a nice convention for methods that belong to the component's lifecycle. The same way `computed` and `methods` have their own describe. I think tests are more readable organized this way ;)

mocha/fixtures/streams/AuthorBar.fixtures.js
christophehenry: Why not? :)

webpack.config.js
```diff
@@ -36,9 +40,15 @@ module.exports = {
     resolve: {
         modules: [
             path.resolve(__dirname, "socialhome"),
+            path.resolve(__dirname, "mocha/fixtures"),
```
jaywink: Now these are probably wrong - and probably unnecessary if the tests are working even if this is wrong ;)

webpack.config.js
```diff
@@ -36,9 +40,15 @@ module.exports = {
...
             path.resolve(__dirname, "node_modules")
         ],
-        alias: {vue: "vue/dist/vue.js",}
+        alias: {
+            "vue$": "vue/dist/vue.esm.js",
+            "bootstrap-vue": "bootstrap-vue/dist/bootstrap-vue.esm.js",
+            "fixtures": path.resolve(__dirname, "mocha/fixtures"),
```
jaywink: Now these are probably wrong - and probably unnecessary if the tests are working even if this is wrong ;)

christophehenry:
I also corrected a few eslint errors this time.

jaywink:
Mucho thanksos! 🍻 ❤️

PR #255, opened 2017-08-13
Add ESLint configuration and fix issues

jaywink:
For new Vue frontend only. Also auto-fix issues.

Note, eslint decided to flatten the .vue <script> tag contents. This is actually recommended here too: https://vuejs.org/v2/guide/single-file-components.html .. So no first level indentation inside the tag.

Ping @christophehenry for review

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,16 @@
+{
+  "extends": [
+    "airbnb-base",
+    "plugin:vue/recommended"
+  ],
+  "rules": {
+    "indent": ["error", 4],
+    "import/extensions": "off",
+    "import/no-extraneous-dependencies": "off",
+    "import/no-unresolved": "off",
+    "max-len": ["error", 120],
+    "object-curly-spacing": ["error", "never"],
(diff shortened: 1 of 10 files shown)


christophehenry:
`<template>` tag is indented. It feels weird that `<script>` and `<style>` are not. I prefer everything is indented.

jaywink:
`<style>` can be indented.

Unfortunately `<script>` is not indented in the Vue single file component examples. And since these are linted as pure `js` files, the scripts must be not indented, otherwise we have to dump the whole linting of indentation. There is no way to tell eslint not to remove the first level indentation in single file vue components only. Unless you can find a way, I tried for a while.

I would prefer to stick to the way the Vue community has found it best to do in this case. Also, we need the linter, otherwise it's a lot of work trying to upkeep code styles. It's painful for contributors (like you) to create code when you have to guess what the style should be like. Linters are a way to document and enforce these styles.

christophehenry:
I think I can live with that :)
We should have a SCSS linter too, I think.

PR #266, opened 2017-08-20
Use websockets

christophehenry:


socialhome/streams/app/components/Stream.vue
```diff
@@ -1,13 +1,13 @@
 <template>
     <div>
         <div class="container-flex">
-            <div id="new-content-container" class="hidden">
-                <a href="" id="new-content-load-link" onclick="return false;">
+            <div v-show="$store.state.stream.hasNewContent" id="new-content-container">
+                <b-button @click.prenvent.stop="onNewContentClick" class="new-content-load-link">
```
jaywink: How does this look visually? Can have screenshot? Current looks like this:

selection_404

A button seems like it could be very blunt looking.

socialhome/streams/app/components/Stream.vue
```diff
@@ -61,6 +61,15 @@ export default Vue.component("stream", {
         onImageLoad() {
             Vue.redrawVueMasonry()
         },
+        onNewContentClick(){
```
jaywink: Space between parenthesis and brace.

socialhome/streams/app/main.js
```diff
@@ -1,6 +1,6 @@
 import Vue from "vue"
-import {VueMasonryPlugin} from "vue-masonry"
 import BootstrapVue from "bootstrap-vue"
+import VueMasonryPlugin from "vue-masonry"
```
jaywink, line 3: Hmm.. How come this started working, when I added the braces I did it because this didn't work. Also, the main readme of the library has braces: https://github.com/shershen08/vue-masonry

socialhome/streams/app/stores/streamStore.js
```diff
@@ -1,31 +1,52 @@
...
+    const state = getState()
+    const store = new Vuex.Store({state, mutations, actions})
+    const wsProtocol = window.location.protocol === "https:" ? "wss" : "ws"
+    const wsPath = `${wsProtocol}://${window.location.host}/ch/streams/${state.stream.streamName}/`
+    const ws = new ReconnectingWebSocket(wsPath)
+
+    ws.onmessage = message => {
+        const data = JSON.parse(message)
+
+        if (data.event === "content") {
```
jaywink: I'm pretty sure this should be `new`.

socialhome/streams/app/stores/streamStore.js
```diff
@@ -1,31 +1,52 @@
 import Vue from "vue"
 import Vuex from "vuex"
-import isNil from "lodash/isNil"
+import ReconnectingWebSocket from "reconnecting-websocket"
 
-// Vue setup
+import getState from "streams/app/stores/streamStore.state"
+import {actions, mutations, stateOperations} from "streams/app/stores/streamStore.operations"
+
+/* This should actually rely in ``main.js``.
```
jaywink: Why should it be in `main.js`, seems like it should be here to me anyway since this is the store? Dunno :D Just thinking what is common in vue. Since everyone uses babel so they must be hitting the same thing and must already have a common way to deal with this.

socialhome/streams/app/stores/streamStore.operations.js
```diff
@@ -0,0 +1,26 @@
...
+    receivedNewContent: "receivedNewContent",
+    newContentAck: "newContentAck",
+}
+
+const mutations = {
+    [stateOperations.receivedNewContent](state, newContentLengh) {
+        state.stream.hasNewContent = true
+        state.stream.newContentLengh += newContentLengh
+    },
+    [stateOperations.newContentAck](state) {
```
jaywink, line 13: What is `newContentAck`?

socialhome/streams/app/stores/streamStore.js
```diff
@@ -1,31 +1,52 @@
...
+    const store = new Vuex.Store({state, mutations, actions})
+    const wsProtocol = window.location.protocol === "https:" ? "wss" : "ws"
+    const wsPath = `${wsProtocol}://${window.location.host}/ch/streams/${state.stream.streamName}/`
+    const ws = new ReconnectingWebSocket(wsPath)
+
+    ws.onmessage = message => {
+        const data = JSON.parse(message)
+
+        if (data.event === "content") {
+            store.dispatch(stateOperations.receivedNewContent, data.contents.length)
```
jaywink: I think the message doesn't contain a `contents`. Did you print it out?

See https://github.com/jaywink/socialhome/blob/67c12ee943c12125c414462c7052d28230e7ffce/socialhome/content/signals.py#L58

jaywink (changes requested):
Looking in the right direction, some comments!

socialhome/streams/app/main.js
christophehenry: This change must have been made a few days ago. With the [version fixing in `package.json`](https://github.com/jaywink/socialhome/pull/266/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L59), it's likely that the library was downgraded.

socialhome/streams/app/stores/streamStore.js
christophehenry: I think all the plugin initialisations should be in only one place. With that problem, we are forced to put the `Vue.use(Vuex)` statement in every store as we can't guaranty in which order the statement are going to be generated. It's bloated.

socialhome/streams/app/stores/streamStore.js
christophehenry: I'm pretty sure I corrected it :'(

socialhome/streams/app/stores/streamStore.operations.js
christophehenry: New content acknowledgment. The user acknowledged new content is available by clicking the button and the state should be reset. Of course, if you think about a better name, I can change.

socialhome/streams/app/components/Stream.vue
christophehenry: I don't know yet. I very admit this is untested :)
I don't know well about HTML accessibility but I feel a button or a link is the way to go for visually disabled people. I should dig into Bootstrap 4 and VueBootstrap to see how components are rendered.

christophehenry:
With latest changes, it looks and work as expected:

screen shot 2017-08-20 at 20 23 25

socialhome/streams/app/stores/streamStore.js
jaywink: At least the official Vuex example repo does this happily in the `store/index.js`, not in application main. I think we should happily do the same and not try put this in main - thus this comment is probably not necessary?

jaywink (approved):

jaywink:
This still WIP? Looks good to me. There is a flaky Python test which failed on build, I restarted it. Need to squash that :(

jaywink:
Oh yeah maybe fix that (IMHO) unnecessary comment first :)

jaywink:
Ah sorry, didn't realize when merging mine it would conflict :(

socialhome/streams/app/stores/streamStore.js
christophehenry: Ok.

socialhome/streams/app/components/Stream.vue
```diff
@@ -33,10 +32,12 @@
 
 <script>
 import Vue from "vue"
+import { mapActions } from "vuex"
```
christophehenry: Uneeded

christophehenry:
Just a few moar tests and it will be all good.

christophehenry:
Oh noes! Random failing tests :,(

socialhome/streams/app/tests/stores/streamStore.tests.js
```diff
@@ -0,0 +1,95 @@
...
+
+        context("when websocket receives a message", () => {
+            it("should dispatch receivedNewContent to store when serveur sends a nex message", (done) => {
+                let mockserver = new Server("wss://localhost:8080/ch/streams/public/")
+                jsdom.reconfigure({url: "https://localhost:8080"})
+                window.context = {streamName: "public"}
+                Sinon.spy(Vuex.Store.prototype, "dispatch")
+                mockserver.on("connection", () => mockserver.send(JSON.stringify({event: "new", id: 4})))
+                newinstance(WebSocket)
+                setTimeout(() => {
```
jaywink, line 41: Any other way to doing the test without a timeout? These end up being really fragile. Like doing the `should` in a callback and then calling `done()`?

package.json
```diff
@@ -52,12 +52,14 @@
     "node": ">=0.8.0"
   },
   "dependencies": {
+    "ReconnectingWebSocket": "https://github.com/joewalnes/reconnecting-websocket.git#1.0.0",
     "axios": "~0.16.2",
     "bootstrap-vue": "~0.20.2",
     "lodash": "~4.17.4",
+    "mock-socket": "^7.0.0",
```
jaywink: `mock-socket` should be a dev dependency only.

jaywink (changes requested):

socialhome/streams/app/tests/stores/streamStore.tests.js
christophehenry: Mocha has a 2s timeout anyway. We must change that. Up to 10s, I'd say. I made a quick search if Mocha can be programmatically configured in `setup.js` yesterday, but I haven't found yet.

christophehenry: But yes, you're right, We must find a better way. Our mock websocket library uses Promises, I think.

package.json
christophehenry: Absolutely. My mistake.

package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "dev": "./node_modules/grunt/bin/grunt dev && ./node_modules/webpack/bin/webpack.js",
     "watch": "./node_modules/grunt/bin/grunt dev && ./node_modules/webpack/bin/webpack.js --watch",
-    "test": "./node_modules/.bin/mocha-webpack --recursive --webpack-config webpack.config.js test --require mocha/setup.js socialhome/streams/app/tests/",
+    "test": "./node_modules/.bin/mocha-webpack --opts mocha test socialhome/**/app/tests/**/*.tests.js",
```
christophehenry: Ah, meh! `--opts mocha/mocha.opts`

mocha/mocha.opts
```diff
@@ -0,0 +1,4 @@
+--recursive
+--webpack-config webpack.config.js
+--require mocha/setup.js
+--timeout 10s
```
jaywink: I don't really like increasing timeouts. Tests should *never* have any reason take more than 2s, which I think you said is the default? If timeouts are happening, something is possibly wrong with the test. Which test is it?

mocha/setup.js
```diff
@@ -1 +1,32 @@
...
+    global.document.destroy = () => {
+        keys.forEach(key => delete global[key])
+    }
+
+    keys.forEach(key => {
+        global[key] = global.window[key]
+    })
+}())
+
+require("chai/register-should")
```
jaywink, line 28: Awesome 👍

mocha/mocha.opts
christophehenry: 10s is a lot, I agree. But why shouldn't the tests have any reason to take more than 2s? Setting up a dom with interaction seems an heavy operation. Espacially with JSDom which is JS. Plus, VueJS is not sync. It has an render loop which caps the render frequency just like the render loop does in a video game. This is why I need to call `Vue.prototype.$nextTick` in some tests. Travis may be slow.

The guilty build is [this one](https://travis-ci.org/jaywink/socialhome/builds/267342961).
[This test, for instance](https://travis-ci.org/jaywink/socialhome/jobs/267342963#L1449) took 1.9 secs. I think 4 secs is a fairly trade off. What do you think?

jaywink: Well, it has a timeout of 1,5s so it's not surprising ;) The real solution is to get rid of the `setTimeout` call, not making sure tests can take longer.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -87,14 +89,14 @@ describe("StreamElement", () => {
                 // Actual thing we are testing - the share
                 target.instance().share()
 
-                moxios.wait(function() {
+                moxios.wait(() => {
```
jaywink, line 92: Sorry I changed two functions to arrow functions but the wrong ones I realized after merging my PR - thanks for doing these :D

jaywink (approved):

christophehenry:
\o/

Issue #206, opened 2017-07-26
Are likes likely?


felixplesoianu:
It would be nice to have likes, favorites, bookmarks or some such, unless it would conflict with the core concepts of Socialhome. Preferably private, but that's not very important.

jaywink:
Actually, both likes and reshares *will* be coming to Socialhome. Likes will either be the traditional "like" or a more varied "reaction" type. The reaction could be a varied type of emoji basically, a bit like on Slack and in a more limited manner Github and Facebook. Probably start with the traditional "boring" like. About the other idea, see #104.

Thanks for reporting!

christophehenry:
I think it may be a good thing to settle down and start thinking carefully about reshares before implementing them. I don't want SH to become the stream-full-of-reshares mess than pollutes diaspora. This raises two points:

1. all reshares of the same post must be disaplayed *only once* in the stream and share the same comments threads,
2. though, sometimes user may want to create a whole new post with its own comments stream and append her own thoughts to the original post.

At the time, I proposed two distinct type of reshares for diaspora: *spread* which does the first, *reshare* which does the second. But I'm not fond of this idea. Stupidely complex. So here is another proposition: the reshare button opens a popup with the possibility to append something to the original post. If the textarea is empty, app just spreads (displayed once, unified comment stream) otherwise, it reshares (brand new post with a link to the original).

jaywink:
I would really like to unify the reshares instead of showing them separately like Diaspora does, so more like for example Twitter which just shows a counter how many people have retweeted. So in theory a big +++ for this.

The problem that comes from here is that Diaspora protocol allows replying to the reshare directly. What do we do with the comments that target the reshare? We can show them in the original maybe. But would that be confusing? Some kind of visual hint would be required to show that "this comment is replying to a reshare of xxx". Threaded comments (technically already possible) would probably have to be supported so that users can reply to the reshare comments.

I would prefer it this way but the UX could be tricky.

The other option is to not care too much and just do it and don't care where the comments are routed inside Diaspora remotes. ActivityPub, for example, does not AFAICT support replying to the reshare. In AP a "reshare" is actually an activity of type `Announce` (confusing, yes) that targets the original object. It isn't an object itself like in Diaspora. And I don't really want to build SH just modelling everything Diaspora does.

felixplesoianu:
[...]

PR #270, opened 2017-08-23
Allow sharing content

jaywink:
Clicking the share counter icon exposes a 'Share' button which when clicked will create a share.

Refs: #206

### DIFF
--- a/socialhome/streams/app/tests/components/StreamElement.tests.js
+++ b/socialhome/streams/app/tests/components/StreamElement.tests.js
@@ -1,5 +1,10 @@
 import "chai/register-should"
 
+import Axios from "axios"
+import moxios from 'moxios'
+import Vue from "vue"
+import {mount} from "avoriaz"
+
 import StreamElement from "streams/app/components/StreamElement.vue"
 import {getStreamElementPropsData} from "../fixtures/StreamElement.fixtures";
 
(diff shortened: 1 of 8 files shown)


socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -37,5 +42,65 @@ describe("StreamElement", () => {
...
+                let propsData = getStreamElementPropsData({isUserAuthenticated: true})
+                let target = mount(StreamElement, {propsData})
+                // Ensure shares are expanded
+                target.instance().expandShares()
+                target.instance().$data.showSharesBox.should.be.true
+
+                // Actual thing we are testing - the share
+                target.instance().share()
+
+                moxios.wait(function() {
```
christophehenry, line 90: Please use arrow function

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -37,5 +42,65 @@ describe("StreamElement", () => {
...
+                    let request = moxios.requests.mostRecent()
+                    request.respondWith({
+                        status: 200,
+                        response: {
+                            status: "ok", content_id: 123,
+                        }
+                    }).then(function () {
```
christophehenry, line 97: Arrow function

socialhome/streams/app/tests/components/StreamElement.tests.js
jaywink: Nice catch. Copied code from interwebz :) Pushed and tests running!

PR #272, opened 2017-08-24
Highlight own shares in stream and allow unsharing

jaywink:
Highlight own shares in stream and allow unsharing.

* Highlight self shared content in grid. Grid share icon will have a slight background colour if the viewer has shared it.
* Allow unsharing content. Clicking the share counter icon exposes an 'Unshare' button (assuming the user has shared the content) which when clicked will remove the share.
* Add share delete endpoint to Content API. This enables removing shares via the API.

peek 2017-08-25 00-31

TODO:
* [x] Add test for has_shares templatetag

Refs: #206


Gruntfile.js
```diff
@@ -182,7 +182,7 @@ module.exports = function (grunt) {
                     "<%= paths.node %>/mocha/mocha.js",
                     "<%= paths.node %>/mocha/mocha.css",
                     "<%= paths.node %>/chai/chai.js",
-                    "<%= paths.node %>/mock-socket/dist/mock-socket.js",
+                    "<%= paths.bower %>/mock-socket/dist/mock-socket.js",
```
christophehenry, line 185: Why? Is it a mistake or there's a reason to add it to Bower again?
*Edit*: Ok, just saw your message on Gitter. Still, can you leave it in node dependencies instead of Bower's?

socialhome/content/viewsets.py
```diff
@@ -47,31 +48,48 @@ class ContentViewSet(mixins.CreateModelMixin, mixins.RetrieveModelMixin, mixins.
...
-    def get_queryset(self):
-        if self.request.user.is_staff:
-            return Content.objects.all()
-        return Content.objects.visible_for_user(self.request.user)
+    def _share(self):
+        content = self.get_object()
+        try:
+            share = content.share(self.request.user.profile)
+        except ValidationError as e:
+            raise exceptions.ValidationError(e.message)
```
christophehenry, line 67: Can't you just do 

```python
except ValidationError:
    raise
```

?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -25,7 +25,9 @@
                 </div>
             </div>
             <div class="ml-auto grid-item-reactions mt-1">
-                <div v-if="showShares" class="item-reaction" @click.stop.prevent="expandShares">
+                <div v-if="showShares" class="item-reaction" v-bind:class="{'item-reaction-shared': getHasShared}"
```
christophehenry: Just do `:class="{'item-reaction-shared': getHasShared}"`. `:` is a shorthand to `v-bind`.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -25,7 +25,9 @@
...
+                     @click.stop.prevent="expandShares"
```
christophehenry: Can you put one attribute per line when multiline?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -88,6 +93,9 @@ export default Vue.component("stream-element", {
         showShares() {
             return this.isUserAuthenticated || this.sharesCount > 0
         },
+        getHasShared() {
```
christophehenry: I'm not fan of the `get`. These are supposed to be attributes and are accessed like ones in the generated Vue code. Maybe `postWasShared`?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -88,6 +93,9 @@ export default Vue.component("stream-element", {
...
+            return this.$data._hasShared
+        },
         getSharesCount() {
```
christophehenry: Idem.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -106,6 +114,20 @@ export default Vue.component("stream-element", {
...
+        unshare() {
+            if (!this.isUserAuthenticated) {
+                console.error("Not logged in")
+                return
+            }
+            this.$http.delete(`/api/content/${this.id}/share/`)
+                .then(() => {
+                    this.$data.showSharesBox = false
+                    this.$data._sharesCount -= 1
+                    this.$data._hasShared = false
```
christophehenry: Maybe it's time to put all data and logic in a store? Not mandatory for this PR. Just throwing a thought.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -12,6 +12,20 @@ describe("StreamElement", () => {
         Sinon.restore()
     })
 
+    describe("template", () => {
+        describe("share reaction", () => {
+            it("has item-reaction-shared if has shared", () => {
+                let propsData = getStreamElementPropsData({hasShared: false})
+                let target = mount(StreamElement, {propsData})
+                target.find(".item-reaction.item-reaction-shared").length.should.eq(0)
```
christophehenry, line 20: `eq`? I don't see that in Chai's documentation. Where does it come from?

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -52,6 +66,24 @@ describe("StreamElement", () => {
                 target.getSharesCount.should.eq(56)
             })
         })
+
+        describe("getHasShared", () => {
+            it("returns has shared status", () => {
+                let propsData = getStreamElementPropsData({hasShared: false})
+                let target = new StreamElement({propsData})
+                target.getHasShared.should.eq(false)
```
christophehenry: `.should.be.false` should work.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -52,6 +66,24 @@ describe("StreamElement", () => {
...
+
+                propsData = getStreamElementPropsData({hasShared: true})
+                target = new StreamElement({propsData})
+                target.getHasShared.should.eq(true)
```
christophehenry: Idem.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -52,6 +66,24 @@ describe("StreamElement", () => {
...
+
+                target.$data._hasShared = false
+                target.getHasShared.should.eq(false)
```
christophehenry: Idem.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -52,6 +66,24 @@ describe("StreamElement", () => {
...
+
+                target.$data._hasShared = true
+                target.getHasShared.should.eq(true)
```
christophehenry: Idem.

socialhome/streams/app/tests/components/StreamElement.tests.js
```diff
@@ -99,6 +132,48 @@ describe("StreamElement", () => {
...
+                // Actual thing we are testing - the unshare
+                target.instance().unshare()
+
+                moxios.wait(() => {
+                    let request = moxios.requests.mostRecent()
+                    request.respondWith({
+                        status: 200,
+                        response: {
+                            status: "ok",
+                        }
```
christophehenry: I was thinking, maybe, simple objects could be one-liners. I'm looking in ESLint's doc for a rule that matches it.

Gruntfile.js
jaywink: I'm pretty sure it would conflict with the newer version we have. I don't want to start fixing the old tests since they'll be deleted as soon as the Vue side is ready. Then we can fully drop bower too.

socialhome/content/viewsets.py
jaywink: We don't want to raise *Django* `ValidationError`, we want to raise *Django REST framework* `ValidationError`. If you look at it closely, they're not the same class. The reason is that if we just pass Django validation error down from the REST viewset, it wont give the same nice error message to the client as the Django REST validation error. That is why we "convert" it here.

socialhome/streams/app/components/StreamElement.vue
jaywink: Ah, nice! 👍

socialhome/streams/app/components/StreamElement.vue
jaywink: I'm not really a fan of how this works at all :( It's annoying to have to specify 3 things just to get one variable to mutate. `postWasShared` doesn't indicate that we're only talking about the current browsing user. I'll read a bit about local state management in Vue, there must be a better way than to have to do this way :/

socialhome/streams/app/components/StreamElement.vue
jaywink: This is just local state for the component instance, it shouldn't belong to the store imho?

socialhome/streams/app/tests/components/StreamElement.tests.js
jaywink: http://chaijs.com/api/bdd/#method_equal - it's a shorter nicer alias for `equal`.

socialhome/streams/app/tests/components/StreamElement.tests.js
jaywink: Ha, must have been tired ;)

socialhome/streams/app/tests/components/StreamElement.tests.js
jaywink: Sure 👍 Btw, have started thinking maybe should revert the change from Airbnb that one liner objects don't have extra spaces, ie:

`{foo: "bar", bar: "foo"}`

Airbnb defines it as:

`{ foo: "bar", bar: "foo" }`

I'm starting to switch my opinion on which is clearer and leaning towards the latter, more commonly used way. Should we switch or keep as is?

socialhome/streams/app/components/StreamElement.vue
jaywink: Well, there doesn't seem to be a way, I guess we talked about this earlier. I think the `_` data was a mistake, it forces to use `computed` to get the actual value in template.

How about one of the three naming conventions, for local data that uses props as initial values. As an example, `hasShared` prop, becomes mutable as:

* `localHasShared`
* `hasSharedLocal`
* `hasShared$`

I kind of like the last one, since it doesn't rely on the word `local` which could have other meanings. If not, I prefer postfixing with `local`, instead of prefixing.

Opinions / other options?

christophehenry: > I'm not really a fan of how this works at all

I know. Unhopefully, Vue components have no constructors like React have. So no way to play with props during construction. There must be a right way. I assume we are not the firsts to run into this use case. I think the whole component must be refactored to use a store anyway. So, for now, `hasShared$` is ok.

socialhome/streams/app/tests/components/StreamElement.tests.js
christophehenry: Ah, yes. I really don't understand this mania of having 36 ways of doing one thing...

socialhome/streams/app/tests/components/StreamElement.tests.js
christophehenry: I'd rather keep as is. To distinguish from one-liner closures. Plus, it's consistent with PEP-8, I think?

Gruntfile.js
christophehenry: Ah, you kept the newer version. Ok, then.

socialhome/content/viewsets.py
christophehenry: Ah, right.

christophehenry (approved):

christophehenry:
Aw... #273 is going to conflict. Stupid me :(

socialhome/streams/app/components/StreamElement.vue
jaywink: Actually, shouldn't the store be an object in the first place, not an array? This way component could just update count of shares in the store instead of maintaining a local mutable copy.. But that is out of scope of this PR, maybe when the API is done refactor the store to be an object so content can be referenced from components by their id.

socialhome/streams/app/tests/components/StreamElement.tests.js
jaywink: Sure!

socialhome/streams/app/components/StreamElement.vue
christophehenry: > This way component could just update count of shares in the store instead of maintaining a local mutable copy..

Yes, that is my idea. But we're still struggling to find the correct way. No need for early optimisations. It works. That's all that matters for now. I'll refactor later. Please merge so I can rebase #273 ;)

PR #273, opened 2017-08-25
Vue JS streams rewrite part 2

christophehenry:
* Fix placement of "new posts label, should be horizontally centered, just below navbar
* Maintains list of ID's for new content waiting on server.
* Content timestamp is a link to the "single content view"
* End of replies section has a "reply" button
* Clicking reply button goes to reply create page


socialhome/streams/app/components/Stream.vue
```diff
@@ -1,45 +1,47 @@
...
+                    :current-browsing-profile-id="$store.state.globalStore.currentBrowsingProfileId"
+                />
             </div>
         </div>
     </div>
 </template>
 
 <script>
 import Vue from "vue"
+import {mapState} from "vuex"
```
christophehenry: Uneeded.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,13 +6,11 @@
...
             :is-user-author="isUserAuthor"
             :is-user-local="isUserLocal"
-            :current-browsing-profile-id="currentBrowsingProfileId"
-            :is-user-authenticated="isUserAuthenticated"
         >
         </author-bar>
         <div class="grid-item-bar d-flex justify-content-start">
             <div class="mt-1">
-                <a :ref="contentUrl" :title="timestamp">
+                <a :href="contentUrl" :title="timestamp">
```
christophehenry: `b-button`

socialhome/streams/app/tests/components/AuthorBar.tests.js
```diff
@@ -7,6 +7,7 @@ import Vue from "vue"
 import VueMasonryPlugin from "vue-masonry"
 
 import AuthorBar from "streams/app/components/AuthorBar.vue"
+import globalStore from "streams/app/stores/globalStore"
```
christophehenry: Unneeded

socialhome/streams/app/tests/fixtures/AuthorBar.fixtures.js
```diff
@@ -1,19 +1,18 @@
...
 const getAuthorBarPropsData = function (args = {}) {
     return _.defaults({}, args, {
-        handle: "",
-        name: "",
-        guid: "",
-        currentBrowsingProfileId: "",
-        homeUrl: "",
-        absoluteUrl: "",
-        imageUrlSmall: "",
+        handle: faker.internet.email(),
```
jaywink: Use `faker.internet.exampleEmail()` to make a safe email that will not accidentally deliver anywhere.

socialhome/streams/app/tests/fixtures/AuthorBar.fixtures.js
```diff
@@ -1,19 +1,18 @@
...
+        name: faker.name.findName(),
+        guid: `${faker.random.number()}`,
```
jaywink: `faker.random.uuid()` is a real fake guid.

socialhome/streams/app/tests/fixtures/AuthorBar.fixtures.js
```diff
@@ -1,19 +1,18 @@
...
+        homeUrl: "https://127.0.0.1",
+        absoluteUrl: "https://127.0.0.1",
+        imageUrlSmall: "https://127.0.0.1",
```
jaywink: `https://127.0.0.1/image.png` would be better.

socialhome/streams/app/tests/fixtures/StreamElementReactionsBar.fixtures.js
```diff
@@ -0,0 +1,15 @@
...
+import faker from "faker"
+
+import {getAuthorBarPropsData} from "streams/app/tests/fixtures/AuthorBar.fixtures";
+
+
+const getStreamElementReactionsBarPropsData = function (args = {}) {
+    return _.defaults({}, args, {
+        id: faker.random.number(),
+        repliesCount: faker.random.number(),
+        sharesCount: faker.random.number(),
```
jaywink: `hasShared - boolean` missing?

socialhome/streams/app/tests/fixtures/StreamElementReactionsBar.fixtures.js
```diff
@@ -0,0 +1,15 @@
+import _ from "lodash/core"
+import faker from "faker"
+
+import {getAuthorBarPropsData} from "streams/app/tests/fixtures/AuthorBar.fixtures";
```
jaywink: This import seems unnecessary.

jaywink (changes requested):

socialhome/streams/app/stores/streamStore.state.js
```diff
@@ -1,24 +1,14 @@
...
-        stream: {
-            hasNewContent: false,
-            newContentLengh: 0,
-            streamName: _get(window, ["context", "streamName"], ""),
-        },
-        isUserAuthenticated: _get(window, ["context", "isUserAuthenticated"], false),
-        showAuthorBar: _get(window, ["context", "showAuthorBar"], false),
-        currentBrowsingProfileId: (_isNumber(_get(window, ["context", "currentBrowsingProfileId"], undefined))
-            ? `${window.context.currentBrowsingProfileId}` : undefined),
+        contentList: _keyBy(_get(window, ["context", "contentList"], []), item => item.id),
```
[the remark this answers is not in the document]
jaywink: Btw, I hope this doesn't mess with the order since the stream must be rendered in the correct order, not by id.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -67,66 +49,21 @@ export default Vue.component("stream-element", {
...
-            }
-            this.$http.delete(`/api/content/${this.id}/share/`)
-                .then(() => {
-                    this.$data.showSharesBox = false
-                    this.sharesCount$ -= 1
-                    this.hasShared$ = false
-                })
-                .catch(err => console.error(err) /* TODO: Proper error handling */)
-        },
+        deleteUrl(){
```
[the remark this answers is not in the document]
christophehenry: It's planned. It's just I fixed the delete button which wasn't working.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,59 +6,41 @@
...
         <div class="grid-item-bar d-flex justify-content-start">
             <div class="mt-1">
-                <a :ref="contentUrl" :title="timestamp">
+                <b-button :href="contentUrl" :title="timestamp" variant="link">
                     {{ humanizedTimestamp }}<span v-if="edited"> {{ editedText }}</span>
-                </a>
+                </b-button>
                 <div v-if="isUserAuthor">
-                    <a :href="updateUrl">
+                    <b-button :href="updateUrl">
```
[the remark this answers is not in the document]
christophehenry: `b-button` generates a link used with attribute `href`. I prefer to use Bootstrap components as much as possible. I'm so concerned about how things can go wrong with CSS.

It does not look the same as before yet. But I need to change the HTML a bit. I broke the design.

socialhome/streams/app/stores/globalStore.js
```diff
@@ -0,0 +1,27 @@
...
+            },
+            isUserAuthenticated: _get(window, ["context", "isUserAuthenticated"], false),
+            currentBrowsingProfileId: (_isNumber(_get(window, ["context", "currentBrowsingProfileId"], undefined))
+                ? `${window.context.currentBrowsingProfileId}` : undefined),
+        },
+    })
+}
+
+const globalStore = newGlobalStore()
+export default globalStore
```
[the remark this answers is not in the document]
christophehenry: `appStore` is fine. I used the term `global` because it's actually a singleton in the code.

socialhome/streams/app/stores/streamStore.js
```diff
@@ -1,25 +1,31 @@
...
 
 Vue.use(Vuex)
 
-function newinstance(WebSocketImpl = ReconnectingWebSocket) {
+function newStreamStore(options) {
     const state = getState()
-    const store = new Vuex.Store({state, mutations, actions})
+    const opts = _defaults({}, {state, mutations, actions}, options)
+    const WebSocketImpl = _get(opts, ["WebSocketImpl"], ReconnectingWebSocket)
+    delete opts.WebSocketImpl
```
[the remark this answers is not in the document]
christophehenry: Test purposes. It has no other use whatsoever. I didn't find a correct way to spy on websocket implementation except by injecting it. After that, I delete it to prevent it from being added to the store's properties. Which is a bit dumb now that I think about it, because it's not used outside tests. I suppose I should add a comment.

socialhome/streams/app/stores/streamStore.state.js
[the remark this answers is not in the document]
christophehenry: > Btw, I hope this doesn't mess with the order since the stream must be rendered in the correct order, not by id.

Good question. Object are not ordered in JS. So my implementation is wrong. I have to use [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

socialhome/streams/app/components/StreamElement.vue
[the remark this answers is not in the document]
jaywink: My point is to me it seems 1) wrong to use a button when you want a link and 2) using a `a` tag will not require any extra styling. So by using the wrong element, we need to add extra directives and/or style to make it look and behave like a normal link.

Additionally, browsers might not interpret a button the same way as a link. For this reason, I think links should be kept as standard HTML specification links, not buttons.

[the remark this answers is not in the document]
christophehenry: > My point is to me it seems 1) wrong to use a button when you want a link

It's not a button. If you add `href` attribute to `b-button`, it generates a link.

socialhome/streams/app/stores/streamStore.state.js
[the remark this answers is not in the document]
christophehenry: Ah, crap. Maps won't work either. Vue can only watch native objects.

socialhome/streams/app/components/StreamElement.vue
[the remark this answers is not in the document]
jaywink: That is like having a class called `Profile` but creating an instance of `Webpage` when the user gives a `url` argument to it. I mean, wtf ;)

Well, I wont fight it, though I think it doesn't make any sense at all to use `b-button` if a normal `a` is *what we really want*.

socialhome/streams/app/stores/streamStore.state.js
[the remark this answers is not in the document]
jaywink: What about storing the list and the objects separately, like;

    contentList: [1,2,3],
    contents: {1: {}, 2:{}, 3:{}},

This way 1) order is guaranteed and 2) objects can be found easily from the components, ensuring we don't need to duplicate store into components as mutable local state.

This allows also inserting directly to the `contentList` from a websocket `new` message. When the object itself appears in `contents`, it will then render it.

socialhome/streams/app/stores/streamStore.js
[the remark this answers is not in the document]
jaywink: Generally of course it would be nice to avoid code that exists only for tests, but if you didn't find any other way to do it then a clear comment please and let's live with it - can always fix it if another way comes up 👍

socialhome/streams/app/components/StreamElement.vue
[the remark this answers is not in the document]
christophehenry: Ok, I went back to an `a`. Using Bootstrap classes modifies the design and design is not the goal here.

socialhome/streams/app/components/StreamElement.vue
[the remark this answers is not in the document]
christophehenry: `contentUrl` [is a bit more complex](https://github.com/jaywink/socialhome/blob/master/socialhome/content/models.py#L154). Not sure it's a good idea.

socialhome/streams/app/components/StreamElementReactionsBar.vue
```diff
@@ -0,0 +1,101 @@
...
+            return this.$store.state.isUserAuthenticated || this.sharesCount$ > 0
+        },
+        replyUrl(){
+            return Urls["content:reply"]({pk: this.id})
+        }
+    },
+    methods: {
+        expandShares() {
+            this.showSharesBox = !this.showSharesBox
+            this.showRepliesBox = this.showSharesBox ? false : this.showRepliesBox
```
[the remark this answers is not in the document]
christophehenry: My own initative. I though it's best to show only one button at a time. It simplifies the UI.

socialhome/streams/app/components/Stream.vue
```diff
@@ -1,30 +1,27 @@
 <template>
     <div>
         <div class="container-flex">
-            <div v-show="$store.state.stream.hasNewContent" class="new-content-container">
+            <div v-show="$store.state.hasNewContent" class="new-content-container">
```
jaywink, line 4: Change this class to an `id` and you don't have to duplicate the CSS. Currently in the non-Vue version this is an `id`.

socialhome/static/sass/streams.scss
```diff
@@ -1,7 +1,8 @@
 @import "variables";
 
 /* New content container */
-#new-content-container {
+#new-content-container,
+.new-content-container {
```
jaywink, line 5: These duplications in this file are unnecessary as other comment re class vs id.

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+    <div class="grid-item-bar d-flex justify-content-start">
+        <slot />
```
jaywink, line 3: What is this? Does it receive the stuff passed into as content when calling this component?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,126 +6,66 @@
...
-        <div class="grid-item-bar d-flex justify-content-start">
+        <reactions-bar
+            :id="id"
+            :replies-count="repliesCount"
+            :shares-count="sharesCount"
+            :has-shared="hasShared"
+        >
             <div class="mt-1">
-                <a :ref="contentUrl" :title="timestamp">
+                <a :href="contentUrl" :title="timestamp" variant="link" class="unstyled-link">
```
jaywink: It's an `a` tag, does it need `variant`?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,126 +6,66 @@
             v-bind="author"
             :is-user-author="isUserAuthor"
             :is-user-local="isUserLocal"
-            :current-browsing-profile-id="currentBrowsingProfileId"
-            :is-user-authenticated="isUserAuthenticated"
         >
         </author-bar>
-        <div class="grid-item-bar d-flex justify-content-start">
+        <reactions-bar
```
jaywink, line 11: How come we're calling a component and then still having content here? What does it mean?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,126 +6,66 @@
...
+        >
             <div class="mt-1">
-                <a :ref="contentUrl" :title="timestamp">
+                <a :href="contentUrl" :title="timestamp" variant="link" class="unstyled-link">
                     {{ humanizedTimestamp }}<span v-if="edited"> {{ editedText }}</span>
                 </a>
-                <div v-if="isUserAuthor">
-                    <a :href="updateUrl">
+                &nbsp;
+                <template v-if="isUserAuthor">
```
jaywink: What does the `template` tag do here, why that instead of `div`?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,126 +6,66 @@
...
+                    <a :href="updateUrl" variant="link">
```
jaywink: It's an `a` tag, does it need `variant`?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -6,126 +6,66 @@
...
                         <i class="fa fa-pencil" title="Update" aria-label="Update"></i>
                     </a>
-                    <a :href="deleteUrl">
+                    &nbsp;
+                    <a :href="deleteUrl" variant="link">
```
jaywink: It's an `a` tag, does it need `variant`?

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
...
+        <div class="ml-auto grid-item-reactions mt-1">
+            <b-button
```
jaywink, line 5: Does the design match the old one? This used to be a `div`.

selection_415

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
...
+                v-if="showShares"
+                :class="{'item-reaction-shared': hasShared$, 'item-reaction': true}"
```
jaywink: I believe you can do `class="item-reaction" :class="{'item-reaction-shared': hasShared$}"` .. ie both `class` and `:class` can exist without having to make it an "if" statement by giving an extra `true` value.

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
...
+                    <span class="item-reaction-counter">{{ repliesCount$ }}</span>
+                </span>
+            </b-button>
+        </div>
+        <div v-if="showSharesBox" class="content-actions">
+            <b-button v-if="hasShared$" variant="secondary" @click.prevent.stop="unshare">Unshare</b-button>
+            <b-button v-if="!hasShared$" variant="secondary" @click.prevent.stop="share">Share</b-button>
+        </div>
+        <div v-if="showRepliesBox">
+            <div class="replies-container"></div>
```
jaywink: This should be outside `<div v-if="showRepliesBox">` before it.

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
...
+            return this.$store.state.isUserAuthenticated || this.sharesCount$ > 0
+        },
+        replyUrl() {
+            return Urls["content:reply"]({pk: this.id})
+        },
+    },
+    methods: {
+        expandShares() {
+            this.showSharesBox = !this.showSharesBox
+            this.showRepliesBox = this.showSharesBox ? false : this.showRepliesBox
```
jaywink: I don't think this is a good idea. Reason: replies. If there are replies, say 10 of them, clicking the share shouldn't make the replies disappear. And if it doesn't, it really doesn't make sense to make the reply button *at the end of the replies*  to disappear.

I would prefer to keep old functionality where the reaction icons just act as toggles and don't depend on each other. It would be different if they were somehow dependent on each other in concept (like in a "you can do one or the other"), but they're not, they're completely different things.

socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -0,0 +1,106 @@
...
+            }
+            this.$http.delete(`/api/content/${this.id}/share/`)
+                .then(() => {
+                    this.showSharesBox = false
+                    this.sharesCount$ -= 1
+                    this.hasShared$ = false
+                })
+                .catch(err => console.error(err) /* TODO: Proper error handling */)
+        },
+        expandComments() {
```
jaywink: One mistake I've made with the Python files - not organizing methods alphabetically. Let's not do that mistake here :)

socialhome/streams/views.py
```diff
@@ -48,14 +48,11 @@ def dump_content(content):
                 "timestamp": content.formatted_timestamp,
                 "humanizedTimestamp": content.humanized_timestamp,
                 "edited": bool(content.edited),
-                "childrenCount": content.reply_count,
+                "repliesCount": content.children_count,
```
jaywink: It's now `content.reply_count` in master 👍

socialhome/streams/app/stores/streamStore.operations.js
```diff
@@ -1,29 +1,46 @@
-const stateOperations = {
+import _forEach from "lodash/forEach"
+
+
+const streamStoreOerations = {
```
jaywink: Typo, should be `streamStoreOperations`.

socialhome/streams/app/components/Stream.vue
```diff
@@ -1,30 +1,27 @@
...
                 <div class="stamped">
-                    <stamped-element></stamped-element>
+                    <stamped-element />
                 </div>
                 <div class="grid-sizer"></div>
                 <div class="gutter-sizer"></div>
                 <stream-element
                     class="grid-item"
-                    v-for="content in $store.state.contentList"
+                    v-for="content in $store.getters.contentList"
```
jaywink, line 19: I'm wondering, won't any change in any content always re-render every single content instead of one of them? Any idea? Also does Vue have to constantly check the getter whether there are new items in the list? I really have no idea how it should work, just wondering out aloud.

We need only one content to re-render when it changes or new appears, otherwise we don't get any benefit from using Vue. Currently the stream re-renders everything on one change.

jaywink (changes requested):
A few change requests and comments, some because I don't know what is going on and the PR is getting to be quite big and it's getting hard to follow. Hope we can clean up and merge soon 👍

socialhome/streams/app/components/Stream.vue
christophehenry: There are two main reasons for using classes here:

1. classes are more efficient than ids for applying CSS (see for instance answer [here](https://stackoverflow.com/questions/7623947/css-id-class-vs-class-performance-which-is-better#7623970)),
2. many of the ids you created are bound to existant JS code, which interfere with Vue rewriting.

jaywink: Ah right, the existing JS 👍

socialhome/streams/app/components/Stream.vue
christophehenry: According to the documentation Vuex' getters are similar to component's computed properties which are cached. As long as the base data don't change, the list don't change. But I suppose the list is entirely recomputed when an element is inserted to or removed from the original list. Which can be heavy, yes.

Good thing is that everything is async in Vue. The view is rerendered the next tick after the property was entirely recomputed so that shouldn't slower the stream in any way.

As for the view re-render, Vue only renders what changed thanks to virtual DOM. This is why list rendering requires [the `id` atribute](https://github.com/jaywink/socialhome/pull/273/files/c8a6be58b6883dfafa3b492096834bf3e1925a25#diff-f6bf769c688863e153cb8f8d589a3803R22), so that only new or removed elements change.

[There's an example close to that in Vuex documentation](https://vuex.vuejs.org/en/getters.html). You can read about the in-place patch strategy [here](https://vuejs.org/v2/guide/list.html#key).

We may face problems with the child component's internal state, though. But as you noticed, our strategy is far from ideal, though I'd consider it out of scope for this PR. I don't think it's time for early optimisation. Especially since we are still discovering Vue's good practices.

socialhome/streams/app/components/ReactionsBar.vue
christophehenry: Yes. It's mainly to keep the current design while leaving the timestamp, delete and edit links, outside the component. It's not pretty, but I think this will change when I start to work on the design.

socialhome/streams/app/components/ReactionsBar.vue
christophehenry: Yes, it does. Styling is attached to `.item-reaction` class.

socialhome/streams/app/components/ReactionsBar.vue
christophehenry: I don't think so but it seems `:class="[{'item-reaction-shared': hasShared$},'item-reaction']"` will work.

socialhome/streams/app/components/ReactionsBar.vue
christophehenry: Alright.

socialhome/streams/app/components/StreamElement.vue
christophehenry: No, it does not. It's... It's... It's a... typo...

socialhome/streams/app/components/StreamElement.vue
christophehenry: Content inserted inside the component tag can be rendered by the component with [the `<slot />` tag](https://vuejs.org/v2/guide/components.html#Single-Slot).

socialhome/streams/app/components/StreamElement.vue
christophehenry: The `template` tag will not be rendered. Only what it contains. It's usefull to do [condiftionnal groups](https://vuejs.org/v2/guide/conditional.html#Conditional-Groups-with-v-if-on-lt-template-gt).

socialhome/streams/app/components/StreamElement.vue
christophehenry: # TYPO ALEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEERT!

socialhome/streams/app/components/StreamElement.vue
christophehenry: Does it look like a typo? Hmm... :D

socialhome/streams/views.py
christophehenry: Breaking changes are evil! :D

socialhome/streams/app/stores/streamStore.operations.js
christophehenry: Hmm... What's a typo? :D

socialhome/streams/app/components/ReactionsBar.vue
jaywink: Sorry forgot to link docs, it does work and IMHO is clearer to define classes as `class` if they don't need to be bind.

> You can have multiple classes toggled by having more fields in the object. In addition, the v-bind:class directive can also co-exist with the plain class attribute. 

https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax

christophehenry: > In addition, the `v-bind:class` directive can also co-exist with the plain `class` attribute.

Indeed.

socialhome/streams/app/components/Stream.vue
jaywink: OK, good enough for me! 👍

config/urls.py
```diff
@@ -9,6 +9,7 @@
 from rest_framework.routers import DefaultRouter
 from rest_framework.schemas import get_schema_view
 from rest_framework_swagger.renderers import OpenAPIRenderer, SwaggerUIRenderer
+from django_js_reverse import views as django_js_reverse_views
```
jaywink: I think you changed this since last review? Please don't use `as` imports unless there is no other way. It makes tracking annoying since the code calls something that isn't the real name what was imported. Even though PyCharm obviously helps, this is a very strong preference for me.

Always import the exact thing you need, unless there are lots of imports then I think it's ok to import a full module.

socialhome/static/sass/streams.scss
```diff
@@ -12,21 +13,24 @@
 }
 
 @media (min-width: $bsMediumDevice) {
-  #new-content-container {
+  #new-content-container,
+  .new-content-container{
```
jaywink: Missing space before {.

socialhome/static/sass/streams.scss
```diff
@@ -12,21 +13,24 @@
...
     font-size: 1.4em;
     top: 38px;
   }
 }
 
 @media (min-width: $bsLargeDevice) {
-  #new-content-container {
+  #new-content-container,
+  .new-content-container{
```
jaywink: Missing space before {.

socialhome/static/sass/streams.scss
```diff
@@ -12,21 +13,24 @@
...
     font-size: 1.2em;
     top: 39px;
   }
 }
 
 @media (min-width: $bsXLargeDevice) {
-  #new-content-container {
+  #new-content-container,
+  .new-content-container{
```
jaywink: Missing space before {.

jaywink (changes requested):
A few easy to fix things left - please don't change anything else to avoid more comments :D Then let's merge!

config/urls.py
christophehenry: > I think you changed this since last review?

No, I didn't. This `as` is quite old.

> Please don't use as imports unless there is no other way.

Even if what's imported as a very generic name?

christophehenry:
Done.

config/urls.py
```diff
@@ -9,6 +9,7 @@
 from rest_framework.routers import DefaultRouter
 from rest_framework.schemas import get_schema_view
 from rest_framework_swagger.renderers import OpenAPIRenderer, SwaggerUIRenderer
+from django_js_reverse import views
```
jaywink: Importing views from a minor third party package is probably not a good idea. I meant import the thing you need only, so "from django_js_reverse.views import urls_js". If you look at other views files, direct import pattern is the regular way.

christophehenry: Ah, right.

christophehenry: Done

jaywink (approved):

jaywink:
Thanks! <3

Issue #202, opened 2017-07-23
Rewrite streams in Vue.js


jaywink:
See high level description at http://socialhome.readthedocs.io/en/latest/roadmap.html#vue-js-app-s

Streams *currently* implemented features below which need to be replicated on the Vue.js side. Tick off when Vue.js side merged to master.

All the items on this list block removing old legacy streams code.

🚫  = blocked by missing API work.
🏆 = does not block releasing the Vue streams as default. Basically stuff not going to implement in legacy streams at all. Create separately tickets maybe if not implemented via this issue.

Stream types:
* [x] Public
* [x] Followed
* [x] Tags
* [x] Profile pinned
* [x] Profile all

Fetching content:
* [x] Stream should use Streams API to fetch correct content for each stream

Content loading (all streams):
* [x] Fetches and renders initially X content items when loading the stream.
* [x] Fetches more content items when scrolling down. Triggers when 5th last item scrolls to view. New content is placed on bottom (order by created date).
  * Content load more last_id must be the lowest "through" content ID in the stream
* [x] Fetches new content when "new posts" label is clicked by user. Content is placed on top (reverse order by created date).
  * [x] Bug: If more than one new post is fetched, the "new posts" label does not disappear. When there is one only, it disappears.
* [x] When "new posts" is clicked, jumps to top of page.

Websockets:
* [x] Creates websocket connection on stream load.
* [x] Shows "new posts" label when server pushes new content id.
* [x] Fix placement of "new posts label, should be horizontally centered, just below navbar
* [x] Maintains list of ID's for new content waiting on server.
* [x] Closes websocket connection when browsing away from stream. (should be automatic, listing for completeness).

Grid items:
* [x] Renders pre-rendered content
* [x] Renders profile small image and name (except if one of the profile streams)
* [x] Profile image or name click opens profile box with handle
* [x] Profile box has "profile", "(remote) home", "follow" and "unfollow" buttons
* [x] Shows content timestamp. Appends "(edited)" if the content is edited.
* [x] Content timestamp is a link to the "single content view"
* [x] Envelop icon shows amount of replies
* [x] Clicking envelop icon fetches replies from the server and shows them
  * [x] 🏆 When fetching replies, show a spinner (see #338)
  * [x] bug: masonry redraw should be done after replies render 
* [x] 🏆 Replies for shares are shown grouped separately with a "reply" button for each share separately
* [x] End of replies section has a "reply" button
* [x] Clicking reply button goes to reply create page
[...]

PR #277, opened 2017-08-28
Add user preference for the new stream refactoring

jaywink:
If enabled, all streams that have a new version in progress will be rendered with the new frontend code based on Vue.js.

Refs: #202


christophehenry:
It may be a good idea to have a more general *enable development features* option. So that it's not purely restricted to the Vue streams port. In the future, other complex tasks can be handled the same way.

PR #280, opened 2017-08-29
Send notification on share and HTML email templates

jaywink:
Additionally, it seems we never sent out a notification if a local user followed a local user. That is now fixed.

Also, ensure Content cache data runs for content that gets replies on shares.

TODO:
* [x] Some extra tests for new code
* [x] Ensure email notification delivered to original author and other participants in original content when a share gets a comment

Refs: #206


socialhome/notifications/templates/notifications/base.html
```diff
@@ -0,0 +1,25 @@
+{% load i18n %}<!doctype html>
```
christophehenry, line 1: Missing a newline here, isn't it?

jaywink: This is actually quite common in the Django world. I guess(?) it is because the HTML spec says a HTML document should start with the doctype. I have no idea if it causes problems or not but having the doctype on a separate line would include empty lines before the doctype. I think it's safer to leave it as is, though I couldn't find a clear answer to whether it could cause problems. Many examples and Django admin templates themselves do it like this.

PR #281, opened 2017-08-31
Remove hardcoded translations from Vue

christophehenry:
- [x] Use VueRouter to match the route and render different stamped elements
- [x] Use Django's gettext library to perform translations
- [x] I can haz tests?

### DIFF
--- a/socialhome/streams/app/components/FollowedStampedElement.vue
+++ b/socialhome/streams/app/components/FollowedStampedElement.vue
@@ -1,0 +1,20 @@
+<template>
+    <base-stamped-element :title="title" :help-text="helpText" />
+</template>
+
+<script>
+import Vue from "vue"
+import "streams/app/components/BaseStampedElement.vue"
+
+
+export default Vue.component("followed-stamped-element", {
+    computed: {
+        title() {
+            return gettext("Followed")
+        },
+        helpText() {
+            return gettext("Content from followed users.")
+        },
+    },
+})
+</script>
--- a/socialhome/streams/app/components/Stream.vue
+++ b/socialhome/streams/app/components/Stream.vue
@@ -33,6 +33,7 @@ import imagesLoaded from "vue-images-loaded"
 import "streams/app/components/StreamElement.vue"
 import PublicStampedElement from "streams/app/components/PublicStampedElement.vue"
 import FollowedStampedElement from "streams/app/components/FollowedStampedElement.vue"
+import TagStampedElement from "streams/app/components/TagStampedElement.vue"
 import {newStreamStore, streamStoreOperations} from "streams/app/stores/streamStore"
 import globalStore from "streams/app/stores/applicationStore"
 
@@ -40,7 +41,7 @@ import globalStore from "streams/app/stores/applicationStore"
 export default Vue.component("stream", {
     store: newStreamStore(),
     directives: {imagesLoaded},
-    components: {FollowedStampedElement, PublicStampedElement},
+    components: {FollowedStampedElement, PublicStampedElement, TagStampedElement},
     data() {
         return {
             masonryOptions: {
@@ -56,10 +57,12 @@ export default Vue.component("stream", {
     },
     computed: {
         stampedElement() {
-            if (this.$store.state.streamName.match(/followed/)) {
+            if (this.$store.state.streamName.match(/^followed/)) {
                 return "FollowedStampedElement"
-            } else if (this.$store.state.streamName.match(/public/)) {
+            } else if (this.$store.state.streamName.match(/^public/)) {
                 return "PublicStampedElement"
+            } else if (this.$store.state.streamName.match(/^tag/)) {
+                return "TagStampedElement"
             } else {
                 console.error(`Unsupported stream name ${this.$store.state.streamName}`)
             }
--- a/socialhome/streams/app/components/Stream.vue
+++ b/socialhome/streams/app/components/Stream.vue
@@ -10,7 +10,7 @@
             </div>
             <div v-images-loaded.on.progress="onImageLoad" v-masonry v-bind="masonryOptions">
                 <div class="stamped">
-                    <stamped-element />
+                    <component :is="stampedElement" />
                 </div>
                 <div class="grid-sizer"></div>
                 <div class="gutter-sizer"></div>
@@ -30,8 +30,9 @@
 <script>
 import Vue from "vue"
 import imagesLoaded from "vue-images-loaded"
-import "streams/app/components/StampedElement.vue"
 import "streams/app/components/StreamElement.vue"
+import PublicStampedElement from "streams/app/components/PublicStampedElement.vue"
+import FollowedStampedElement from "streams/app/components/FollowedStampedElement.vue"
 import {newStreamStore, streamStoreOperations} from "streams/app/stores/streamStore"
 import globalStore from "streams/app/stores/applicationStore"
 
@@ -39,6 +40,7 @@ import globalStore from "streams/app/stores/applicationStore"
 export default Vue.component("stream", {
     store: newStreamStore(),
     directives: {imagesLoaded},
+    components: {FollowedStampedElement, PublicStampedElement},
     data() {
         return {
             masonryOptions: {
@@ -51,6 +53,17 @@ export default Vue.component("stream", {
                 stagger: 0,
             },
         }
+    },
+    computed: {
+        stampedElement() {
+            if (this.$store.state.streamName.match(/followed/)) {
+                return "FollowedStampedElement"
+            } else if (this.$store.state.streamName.match(/public/)) {
+                return "PublicStampedElement"
+            } else {
+                console.error(`Unsupported stream name ${this.$store.state.streamName}`)
+            }
+        },
     },
     methods: {
         onImageLoad() {
--- a/socialhome/streams/app/components/TagStampedElement.vue
+++ b/socialhome/streams/app/components/TagStampedElement.vue
@@ -1,0 +1,20 @@
+<template>
+    <base-stamped-element :title="title" :help-text="helpText"/>
+</template>
+
+<script>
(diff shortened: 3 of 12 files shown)


socialhome/streams/app/components/Stream.vue
```diff
@@ -63,6 +62,9 @@ export default Vue.component("stream", {
     beforeDestroy() {
         this.$store.$websocket.close()
     },
+    created(){
+        console.log(this.$route.path)
+    }
```
christophehenry: Nope!

socialhome/streams/app/components/FollowedStampedElement.vue
```diff
@@ -0,0 +1,20 @@
...
+import GenericStampedElement from "streams/app/components/GenericStampedElement.vue"
+
+
+export default Vue.component("followed-stamped-element", {
+    computed: {
+        title() {
+            return gettext("Followed")
+        },
+        helpText() {
+            return gettext("Contains public content from around the network.")
```
christophehenry: Should be `"Content from followed users."`

jaywink:
With this solution, one general concern I have is how streams mapped to root `/` will be handled. Can `vue-router` handle complex "if/else" logic for a single route? If it can, we *could* export the user preferences controlling what is rendered and then use those preferences to decide like in Django [here](https://github.com/jaywink/socialhome/blob/f048f0adc5c07ce48e562351ae78dabf799a2622/socialhome/views.py#L16).

But to be honest, you will need to base what stream is rendered (or what to fetch via API) anyway based on the stream name - or the exported preferences.

To me, it feels it would be simpler just looking at "what is stream name" and then rendering the right stamped element component based on that. It's much simpler and the decision which sub component to render is in the right place, the `StreamComponent`. Doing this way feels like a bit of a hack to use the router to decide on which sub component is rendered. I understand the case for a router if we actually *do routing* inside Vue. But this is more of a "if route this, render this component  when rendering the stream, which isn't itself routed".

So, TL/DR, I support the idea of different components for the different stream stamped elements (a must for profile stamped element, it's quite complex), but I feel there is a simpler way to decide which one to render. A stream is *not* mapped to a single URL in Socialhome, so you can't absolutely decide that Followed stream is always `/followed`.

christophehenry:
With this solution, one general concern I have is how streams mapped to root `/` will be handled. Can `vue-router` handle complex "if/else" logic for a single route? If it can, we could export the user preferences controlling what is rendered and then use those preferences to decide like in Django [here](https://github.com/jaywink/socialhome/blob/f048f0adc5c07ce48e562351ae78dabf799a2622/socialhome/views.py#L16).

I could specify multiple components [here](https://github.com/jaywink/socialhome/pull/281/files#diff-14c22c262c80bfa58a8ccc08463d9ae2R10) and change [the name](https://github.com/jaywink/socialhome/pull/281/files#diff-f6bf769c688863e153cb8f8d589a3803R13) using a data property in the component.

> To me, it feels it would be simpler just looking at "what is stream name" and then rendering the right stamped element component based on that.

When the time of vueifying profiles pages come, I don't want to do things [like that](https://github.com/jaywink/socialhome/blob/master/socialhome/streams/app/stores/streamStore.state.js#L18). Having a stream name variable feels wrong anyway.

>  I understand the case for a router if we actually _do routing_ inside Vue.

Oh, we will do routing very soon. I intend to render any stream on front-end anyway so we can use caching of data client-side and epically reduce network load.

socialhome/streams/app/components/GenericStampedElement.vue
```diff
@@ -15,5 +15,11 @@ import Vue from "vue"
 import store from "streams/app/stores/applicationStore"
 
 
-export default Vue.component("stamped-element", {store})
+export default Vue.component("generic-stamped-element", {
+    store,
```
christophehenry: Useless, now

socialhome/streams/app/components/FollowedStampedElement.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+    <generic-stamped-element :title="title" :help-text="helpText" />
+</template>
+
+<script>
+import Vue from "vue"
+import GenericStampedElement from "streams/app/components/GenericStampedElement.vue"
```
jaywink: `Generic` kind of keeps colliding with me every time I look at this. Can we use `BaseStampedElement`? "Base" is a more common word to use in these cases for example with Django templates or classes.

socialhome/streams/app/components/Stream.vue
```diff
@@ -52,6 +53,18 @@ export default Vue.component("stream", {
             },
         }
     },
+    computed: {
+        stampedElement() {
```
jaywink, line 59: A simple test for this computed function would be nice 👍

jaywink (commented):
Looks awesome now! A few comments.

christophehenry:
Done

socialhome/streams/app/components/Stream.vue
```diff
@@ -52,6 +55,19 @@ export default Vue.component("stream", {
...
+            if (this.$store.state.streamName.match(/followed/)) {
```
jaywink: You must use `starts with` kind of check OR do a split and take the first element. For example currently tag streams have the tag name slug in the stream name, so #followed would get the followed stamped element. In a way this is bad and unnecessary, but it's the way it is. And the "spec" for stream names is basically `<stream_type>__<id or other identifier>`. The first two underscores end the stream type definition which should be used.

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -1,13 +1,17 @@
 <template>
     <div>
-        <div v-html="htmlSafe"></div>
+        <div v-if="nsfw" class="card card-block text-center nsfw-shield">
+            <b-button href="#" variant="link" @click.stop.prevent="toggleNsfw">
+                {{ toggleNsfwBtnText }}
+            </b-button>
+        </div>
+        <div v-if="showNsfw" v-html="htmlSafe"></div>
```
jaywink: This is incorrect. The NSFW shield is already in the rendered contents in Socialhome. There is an `nsfw` class added to all images in the content if a tag is found. This activates the NSFW shield. What you need in the JS side is a trigger to show it on click. There should also be a trigger then to return the nsfw shield. Try it to see how it works.

We don't want to hide the whole content, just images. Diaspora NSFW shield always annoyed me because you can't see any context to the post before clicking it.

socialhome/streams/app/components/TagStampedElement.vue
```diff
@@ -0,0 +1,20 @@
...
+import "streams/app/components/BaseStampedElement.vue"
+
+
+export default Vue.component("tag-stamped-element", {
+    computed: {
+        title() {
+            return `#${this.$store.state.tagName}`
+        },
+        helpText() {
+            return interpolate(gettext("All content tagged %s"), [`#${this.$store.state.tagName}`])
```
jaywink: Used to be just the tag name, but if you want to put some more text that is fine. Make sure it ends in a period though ;) Also, I think it should be "All content tagged *with* %s".

socialhome/streams/app/tests/components/Stream.tests.js
```diff
@@ -14,6 +16,28 @@ describe("Stream", () => {
...
+                let target = mount(Stream, {})
+                target.instance().$store.state.streamName = "followed"
+                target.instance().stampedElement.should.eq("FollowedStampedElement")
+            })
+            it("should render the PublicStampedElement when stream name is 'public'", () => {
+                let target = mount(Stream, {})
+                target.instance().$store.state.streamName = "public"
+                target.instance().stampedElement.should.eq("PublicStampedElement")
+            })
+            it("should display an error when stream name is unknown", () => {
```
jaywink: Missing test for tag stream.

socialhome/streams/app/tests/components/Stream.tests.js
```diff
@@ -14,6 +16,28 @@ describe("Stream", () => {
...
+                let target = mount(Stream, {})
+                Sinon.spy(console, "error")
+                target.instance().$store.state.streamName = "Yolo stream"
```
jaywink: Should we make a Yolo stream??

jaywink (changes requested):
Unfortunately the NSFW behaviour does not mirror what exists now.

socialhome/streams/app/components/Stream.vue
christophehenry: `/^followed/` might be enough, then.

socialhome/streams/app/components/StreamElement.vue
christophehenry: Hmm... I don't like that. It's not the Vue way. It will require messing with selectors :/

jaywink: Can't bend features to code.. Code must bend to features ;)

socialhome/streams/app/tests/components/Stream.tests.js
christophehenry: Most definitely!

christophehenry:
Ready for another review

socialhome/streams/app/components/TagStampedElement.vue
```diff
@@ -0,0 +1,20 @@
...
+import "streams/app/components/BaseStampedElement.vue"
+
+
+export default Vue.component("tag-stamped-element", {
+    computed: {
+        title() {
+            return `#${this.$store.state.tagName}`
+        },
+        helpText() {
+            return interpolate(gettext("All content tagged with %s"), [`#${this.$store.state.tagName}`])
```
jaywink: Missing period from sentence 👍

jaywink (changes requested):
One missing period and then merge! 💃

jaywink (approved):

jaywink:
Thanks!

PR #286, opened 2017-09-02
Remove properties passing down to the stream child components and use store inheritance

christophehenry:
- [x] Remove properties passing down to the stream child components and use store inheritance
- [x] I can haz tests?


socialhome/streams/app/components/StreamElement.vue
```diff
@@ -41,31 +30,27 @@ import store from "streams/app/stores/applicationStore"
...
         deleteUrl() {
-            return Urls["content:delete"]({pk: this.id})
+            return Urls["content:delete"]({pk: this.contentId})
+        },
+        editedText() {
+            return this.content.edited ? `${this.content.humanizedTimestamp} (edited)` : this.content.humanizedTimestamp
+        },
+        isUserAuthor(){
+            return this.$store.state.applicationStore.currentBrowsingProfileId === this.content.author.guid
+        },
```
christophehenry: Can probably be removed (and doesn't work anyway)

christophehenry:
Ready for review.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -94,17 +93,17 @@ export default Vue.component("author-bar", {
                 console.error("Not logged in")
                 return
             }
-            this.$http.post(`/api/profiles/${this.currentBrowsingProfileId}/remove_follower/`, {guid: this.guid})
-                .then(() => { this.$data._isUserFollowingAuthor = false })
+            this.$http.post(`/api/profiles/${this.currentBrowsingProfileId}/remove_follower/`, {guid: this.author.guid})
+                .then(() => { this.author.isUserFollowingAuthor = false })
```
jaywink, line 97: This works and updates the store value even though `this.author` is a `computed` value?

socialhome/streams/app/components/StreamElement.vue
```diff
@@ -1,25 +1,14 @@
...
-            :has-shared="hasShared"
-        >
+        <div v-html="content.htmlSafe"></div>
+        <author-bar v-if="showAuthorBar" :content-id="contentId" />
+        <reactions-bar :content-id="contentId">
             <div class="mt-1">
-                <a :href="contentUrl" :title="timestamp" class="unstyled-link">
-                    {{ humanizedTimestamp }}<span v-if="edited"> {{ editedText }}</span>
+                <a :href="content.contentUrl" :title="content.timestamp" class="unstyled-link">
+                    {{ editedText }}
```
jaywink: `timestampText` would be better considering the refactoring 👍

jaywink (changes requested):
One minor comment in addition to the question - but then I assume it does work because you've tested it ;)

socialhome/streams/app/components/AuthorBar.vue
christophehenry: Yep :)

jaywink (approved):

jaywink:
Thanks! <3<3

PR #288, opened 2017-09-04
Streams URL changes

jaywink:
* All streams will now be under `/streams/` for a cleaner URL layout. So for example `/public/` is now `/streams/public/`.
* Tag stream URL has been changed from `/streams/tags/<tag>/` to `/streams/tag/<tag>/`. This small change allows us to later map `/stream/tags/` to the tags the user is following.

Since lots of old content will point to the old URL's, there will be support for the legacy URL's until they are needed for something else in the future.

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -33,7 +33,9 @@ urlpatterns = [
     url(r"", include("socialhome.federate.urls", namespace="federate")),
 
     # Streams
-    url(r"", include("socialhome.streams.urls", namespace="streams")),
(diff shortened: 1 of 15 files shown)


christophehenry:
Are users also concerned by this change, *i.e* will become `/streams/u/<userid>`?

jaywink:
No, those are `profile` URL's. Internally the content of those urls just happen to be streams! :) Just like a content detail page will be a stream of replies.

PR #316, opened 2017-10-05
Alpine dependencies installation script

0rzech:


install_alpine_dependencies.sh
```diff
@@ -0,0 +1,58 @@
...
+  curl \
+  llvm4 \
+  `# federation` \
+  libxml2-dev \
+  libxslt-dev \
+  python3-dev \
+  `# socialhome` \
+  git \
+  linux-headers \
+  nodejs \
```
jaywink, line 49: Which Node is this btw? Currently I think anything goes (though I have only tested 4+), but to compile the Vue.js frontend bundle, Node 6 seems to be required. I have been meaning to add this already to the documentation, even though the Vue frontends are not ready yet.

jaywink (commented):
Just one comment re node version!

install_alpine_dependencies.sh
0rzech: It's [this](https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.6) one.

jaywink: Thanks that is sufficient 👍

Issue #111, opened 2017-03-07
Don't load NSFW images before user clicks to show them


jaywink:
Currently browser loads but hides them. Load on click.

christophehenry:
There are a bunh of plugins to lazy-load images. Can add this when streams are rewritten in Vue.

jaywink:
This is fixed in Vue streams with #330. They will be default soon so no need to fix this with the legacy streams.

PR #330, opened 2017-10-17
Add NSFW shield

christophehenry:
Ok, here is my proposition for a NSFW shield:

- hidden

screenshot-2017-10-17 - socialhome
- shown

screenshot-2017-10-17 - socialhome 1

When hidden, it only shows the tags. BTW, for an unknown reason, the tags `field` of the API does not gives the correct tags. In this example, it's an array containing `3`.

The current implementation solves #111 as the HTML element is not even redered until the user clicks the button to show.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,6 +27,10 @@ Changed
 
   This is a more standard file name for environment variables than the previous ``env.local``. For now we'll still load from the old file too, but a warning will be displayed to rename the file.
 
+* **Breaking change**
+
+  Content API now returns Content list of tags as *name of tag*, not ID as before. The name does not contain the character "#".
+
 Fixed
 .....
 
--- a/socialhome/content/serializers.py
+++ b/socialhome/content/serializers.py
@@ -1,6 +1,6 @@
 from enumfields.drf import EnumField
 from rest_framework.fields import SerializerMethodField
-from rest_framework.serializers import ModelSerializer
+from rest_framework.serializers import ModelSerializer, SlugRelatedField
 
 from socialhome.content.enums import ContentType
 from socialhome.content.models import Content
@@ -14,6 +14,7 @@ class ContentSerializer(ModelSerializer):
     user_following_author = SerializerMethodField()
     user_is_author = SerializerMethodField()
     user_has_shared = SerializerMethodField()
+    tags = SlugRelatedField(many=True, read_only=True, slug_field="name")
     through = SerializerMethodField()
     visibility = EnumField(Visibility, lenient=True, ints_as_names=True)
 
@@ -26,6 +27,7 @@ class ContentSerializer(ModelSerializer):
             "guid",
             "humanized_timestamp",
             "id",
+            "is_nsfw",
             "local",
             "order",
             "parent",
@@ -52,6 +54,7 @@ class ContentSerializer(ModelSerializer):
             "guid",
             "humanized_timestamp",
             "id",
+            "is_nsfw",
             "local",
             "parent",
             "remote_created",
--- a/socialhome/content/tests/test_serializers.py
+++ b/socialhome/content/tests/test_serializers.py
@@ -3,7 +3,7 @@ from unittest.mock import Mock
 from django.contrib.auth.models import AnonymousUser
 
 from socialhome.content.serializers import ContentSerializer
-from socialhome.content.tests.factories import PublicContentFactory
+from socialhome.content.tests.factories import PublicContentFactory, TagFactory
 from socialhome.tests.utils import SocialhomeTestCase
 
 
@@ -89,3 +89,15 @@ class ContentSerializerTestCase(SocialhomeTestCase):
         self.content.share(self.profile)
         serializer = ContentSerializer(context={"request": Mock(user=self.user)})
         self.assertTrue(serializer.get_user_has_shared(self.content))
+
+    def test_tags_if_no_tag(self):
+        self.content.tags.clear()
+        serializer = ContentSerializer(self.content, context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.data["tags"], [])
+
+    def test_tags_with_tag(self):
+        tag = TagFactory(name="yolo")
+        self.content.tags.clear()
+        self.content.tags.add(tag)
+        serializer = ContentSerializer(self.content, context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.data["tags"], ["yolo"])
--- a/socialhome/streams/app/components/NsfwShield.vue
+++ b/socialhome/streams/app/components/NsfwShield.vue
@@ -1,0 +1,52 @@
+<template>
+    <div>
+        <b-button
+            v-if="!showNsfwContent"
+            @click.stop.prevent="toggleNsfwShield"
+            class="text-center nsfw-shield card card-block"
+            href="#"
+            variant="link"
+        >
+            {{ nsfwBtnText }}
+        </b-button>
+        <div v-else class="text-center nsfw-shield">
+            <a @click.stop.prevent="toggleNsfwShield" href="#">{{ nsfwBtnText }}</a>
+        </div>
+        <div v-show="!showNsfwContent">
+            <b-button v-for="tag in tags" :href="getTagUrl(tag)" variant="link">#{{tag}}</b-button>
+        </div>
+        <slot v-if="showNsfwContent" />
+    </div>
+</template>
+
+<script>
+import Vue from "vue"
+
+
+export default Vue.component("nsfw-shield", {
+    props: {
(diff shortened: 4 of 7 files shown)


jaywink:
Awesome! A few notes on the design:

- The tags (when NSFW shield on) should be normally rendered tags (text), not pills. We don't use pills to render tags in the content normally, so it feels weird to have tags rendered in two different ways. They should probably be a link to the tag stream too, like tags normally are. That's kinda optional, though if not, then they shouldn't look like links. If linkifying the tags, make sure to lowercase the whole tag name before making it a link. Otherwise any tag in the system should be "link safe".

- When the shield is off, the "Hide NSFW content" should be more hidden away normal text centered before the rendered content, for example like it is now (though 'content', not 'image', obviously):

selection_431

Open to alternatives regarding this, but definitely the box should not be reused for the hiding. The point of the box is to give a clear action to the common thing a user would do with this content. The hiding is more of a secondary action and should be less visible. What do you think?

About the API, yes, the list of tags is currently the ID of the tag, not the name. I think that should probably be changed. You can just pull whatever is in the array for this PR. I can make the API change unless you want to read a bit more into Django REST framework? Basically need to tell it to serialize not the ID but the name of the foreign key relation.

socialhome/streams/app/components/NsfwShield.vue
```diff
@@ -0,0 +1,51 @@
...
+        nsfwBtnText() {
+            return this.showNsfwContent ? gettext("Hide NSFW content") : gettext("Show NSFW content")
+        },
+    },
+    methods: {
+        toggleNsfwShield() {
+            this.showNsfwContent = !this.showNsfwContent
+        },
+        getTagUrl(name) {
+            return Urls["streams:tag"]({name})
```
jaywink, line 45: Great, you already did the linkifying. Actually ignore my point about lowercasing - I realized we save all tags lowercase in the `Tag` model when ripping them out of content. So this method is fine.

christophehenry:
> The tags (when NSFW shield on) should be normally rendered tags (text), not pills. We don't use pills to render tags in the content normally, so it feels weird to have tags rendered in two different ways. 

I was thinking that it's actually a good idea to highlight tags for all posts just below and render tags in post more discretely. It could be easier to read for posts with lots of tags. We could, for instance render tags in text in a slightly lighter grey than text and highlight them as pills.

It would also have another pro: discouraging people to put too many tags in a post to avoid tag spamming which is a pain in diaspora.

> The hiding is more of a secondary action and should be less visible. What do you think?

No strong opposition on that. It's easy to implement and makes sense.

> I can make the API change unless you want to read a bit more into Django REST framework?

Yup. I need to level up on this. I prefer taking care of this.

jaywink:
We can discuss the highlighting of tags in other than "text with link" in some other future time. Though, personally, I do prefer pure text tag links. It's a standard wherever you look. It makes them stand out but not stand out too much. It doesn't ruin the text flow when reading posts with tags. It doesn't ruin formatting. But always open to mockups which we can discuss :)

However, I think things like that are out of scope of this PR and the tags should have the same rendering we use in content itself.

> It would also have another pro: discouraging people to put too many tags in a post to avoid tag spamming which is a pain in diaspora.

Most content comes from diaspora, or in the future from Mastodon, tho :) Unless Socialhome nodes explode in number into the thousands.

> Yup. I need to level up on this. I prefer taking care of this.

\o/ REST framework is fun and pretty well documented.

christophehenry:
Indeed, you're right, styling out of scope. Let's stick to pure text links.

christophehenry:
I'm having troubles testing things. Running `py.test`, I get:

```
Got an error creating the test database: permission denied to create database

Got an error recreating the test database: database "test_socialhome" does not exist
```

socialhome/content/tests/test_serializers.py
```diff
@@ -89,3 +89,15 @@ def test_user_has_shared_true_if_shared(self):
...
+        self.content.tags.clear()
+        serializer = ContentSerializer(context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.get_tags(self.content), [])
+
+    def test_tags_with_tag(self):
+        tag = TagFactory(name="yolo")
+        self.content.tags.clear()
+        self.content.update(tags=[tag])
+        serializer = ContentSerializer(context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.get_tags(self.content), [])
```
christophehenry: Halp!

jaywink:
> Got an error creating the test database: permission denied to create database

You need to give the user rights to create databases, for example:

    sudo -u postgres psql -c "ALTER USER socialhome CREATEDB;"

Seems this is missing from the development docs, sorry, let me know if it was the issue and will add!

socialhome/content/tests/test_serializers.py
```diff
@@ -89,3 +89,15 @@ def test_user_has_shared_true_if_shared(self):
...
+
+    def test_tags_if_no_tag(self):
+        self.content.tags.clear()
+        serializer = ContentSerializer(context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.get_tags(self.content), [])
+
+    def test_tags_with_tag(self):
+        tag = TagFactory(name="yolo")
+        self.content.tags.clear()
+        self.content.update(tags=[tag])
```
jaywink: This should be `self.content.tags.add(tag)`.

socialhome/content/serializers.py
```diff
@@ -91,3 +94,6 @@ def get_user_has_shared(self, obj):
         if not request:
             return False
         return Content.has_shared(obj.id, request.user.profile.id) if hasattr(request.user, "profile") else False
+
+    def get_tags(self, obj):
```
jaywink: While this works, it's inefficient. REST framework has a thing to do this, it's the [SlugRelatedField](http://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield). Using that exactly as the example there (with `read_only` and `many`, field name `name`) should work nicely and be well optimized.

socialhome/content/serializers.py
```diff
@@ -14,6 +14,7 @@ class ContentSerializer(ModelSerializer):
     user_following_author = SerializerMethodField()
     user_is_author = SerializerMethodField()
     user_has_shared = SerializerMethodField()
+    tags = SlugRelatedField(many=True, read_only=True, slug_field="tags")
```
jaywink: `slug_field` should be `name`.

socialhome/content/serializers.py
jaywink: This method is not necessary when using `SlugRelatedField`.

socialhome/content/tests/test_serializers.py
```diff
@@ -89,3 +89,15 @@ def test_user_has_shared_true_if_shared(self):
         self.content.share(self.profile)
         serializer = ContentSerializer(context={"request": Mock(user=self.user)})
         self.assertTrue(serializer.get_user_has_shared(self.content))
+
+    def test_tags_if_no_tag(self):
+        self.content.tags.clear()
+        serializer = ContentSerializer(context={"request": Mock(user=self.user)})
+        self.assertEquals(serializer.get_tags(self.content), [])
```
jaywink: These tests should now check the serialized content object and what `tags` list contains. You will find the list as `serializer.data["tags"]`.

jaywink (changes requested):
A few things to fix re serializer.

Also, please add a changelog entry to the "Changed" section in the current unreleased version. Something like:

    **Breaking change**. Content API now returns Content list of tags as *name of tag*, not ID as before. The name does not contain the character "#".

socialhome/content/serializers.py
```diff
@@ -1,9 +1,9 @@
 from enumfields.drf import EnumField
 from rest_framework.fields import SerializerMethodField
-from rest_framework.serializers import ModelSerializer
+from rest_framework.serializers import ModelSerializer, SlugRelatedField
 
 from socialhome.content.enums import ContentType
-from socialhome.content.models import Content
+from socialhome.content.models import Content, Tag
```
jaywink: Tag import is unnecessary.

jaywink (changes requested):
One more comment. Also, that changelog entry would be nice 👍  Saves an extra follow up commit.

docs/changelog.rst
```diff
@@ -27,6 +27,10 @@ Changed
 
   This is a more standard file name for environment variables than the previous ``env.local``. For now we'll still load from the old file too, but a warning will be displayed to rename the file.
 
+* **Breaking change**
```
jaywink, line 30: The changelog is a list inside each header, but I'll fix this in a separate commit, noticed one other thing to add too 👍

jaywink (approved):

jaywink:
Thanks! <3

PR #334, opened 2017-10-18
First attempt at a docker development environment

ghost:
This is the promised docker development environment


.env.example
```diff
@@ -11,3 +11,9 @@ SOCIALHOME_DOMAIN=127.0.0.1:8000
 SOCIALHOME_HTTPS=False
 
 MOCHA_TESTS=True
+
+# Set these variables for Docker Postfix (be sure the DATABASE_URL matches)
```
jaywink: `Postfix` -> `PostgreSQL`.

I'm also wondering why do we need these here at all? Why would anyone want to change these for a development database? IMHO, they should just be set in `docker-compose.yml`. If someone *wants* to change these, they can change that file - it's a local copy anyway not part of the repo. If possible, it would be nice to not have extra optional stuff in `.env.example` - it already has a few unnecessary things that should be cleaned out.

docker/dev/Dockerfile.django
```diff
@@ -0,0 +1,32 @@
+# We use the official Python images. This may need to get locked in, or back to Ubuntu at some point
```
jaywink: Maybe this commetn can be removed if the challenge has already been solved :)

docker/dev/Dockerfile.django
```diff
@@ -0,0 +1,32 @@
...
+FROM python:3
+
+RUN apt-get update
+RUN apt-get install -y lib32z1-dev lib32z1 libpq-dev
```
jaywink: I'm kind of surprised `libxml2-dev libxslt-dev` are not needed. Those are direct dependencies for `lxml` to be built successfully. Maybe the Python3 image has them already?

docker/dev/Dockerfile.npm
```diff
@@ -0,0 +1,17 @@
+# We use the official Python images. This may need to get locked in, or back to Ubuntu at some point
```
jaywink: Redundant comment

docker/dev/Dockerfile.npm
```diff
@@ -0,0 +1,17 @@
...
+FROM node
+
+RUN mkdir /code
+
+WORKDIR /code
+COPY . .
+
+RUN npm install
+RUN node_modules/.bin/bower --allow-root install
+RUN node_modules/.bin/grunt build
```
jaywink: `grunt dev` when doing development - this ensures the scripts are not minified, helping debug problems.

~~Also, `npm run dev` is missing. This builds the webpack bundle. Failing to do so will crash the UI when trying to access a Vue view.~~

docker/dev/docker-entrypoint.sh.django
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+#
+# Socialhome Docker Development entrpypoint
```
jaywink: Typo in "entrypoint"

docs/development.rst
```diff
@@ -63,11 +63,11 @@ To watch files and build bundles automatically, use this.
 Configure
 .........
 
-Configuration is done via environment variables. For the meaning of them, look them up under files in ``config/settings``. Values in the file ``.env`` will be used automatically.
+Configuration is done via environment variables. For the meaning of them, look them up under files in ``config/settings``. Values in ``env.local`` will be used automatically.
```
jaywink: This is reverting my previous commit. I think you somehow managed to copy the old file with your changes, thus reverting some other stuff I committed since. This is why it is important to develop in branches, not in the master branch.

docs/development.rst
```diff
@@ -63,11 +63,11 @@ To watch files and build bundles automatically, use this.
...
 
 ::
 
-    cp .env.example .env
+    cp env.example env.local
```
jaywink: Reverting master changes here too.

docs/development.rst
```diff
@@ -148,18 +148,6 @@ Execute the following to run the new frontend JavaScript tests.
 
     npm run test
 
-API routes
```
jaywink, line 151: Reverting master changes here too.

docs/development.rst
```diff
@@ -203,6 +191,42 @@ Then execute the following and copy the markdown version for pasting to GitHub r
 
 After the release commit has been pushed and a release has been tagged, set a development version in the same above files. This is basically the next minor release postfixed by ``-dev``.
 
+Developing with Docker
+------------------------
+
+If you choose, you may develop Socialhome using Docker, rather than installing Postfix and Redis manually on your computer.
```
jaywink: `Postfix` -> `PostgreSQL`.

Maybe someday we *will* have a postfix dependency to allow Socialhome to act as a mail server ;)

docs/development.rst
```diff
@@ -203,6 +191,42 @@ Then execute the following and copy the markdown version for pasting to GitHub r
...
+And then the steps you would normally do, but throught he django image, ala:
+
+``docker-compose run django manage migrate``
+and
+``docker-compose run django manage createsuperuser``
+
+And then just
+
+``docker-compose up``
+
```
jaywink: I wonder if it would make sense to also mention where the UI is exposed to, ie what to type in the browser to reach the running instance? Might help not so experienced developers. It's not apparent to me at least where I should look.

docs/development.rst
```diff
@@ -203,6 +191,42 @@ Then execute the following and copy the markdown version for pasting to GitHub r
...
+..................
+
+This guide assumes you are running Docker on a GNU/Linux based system such as Ubuntu, Debian or Fedora Linux. It may be possible to run this on other platforms where Docker is supported, but those are untested.
+
+The docker development installation was tested on Docker version 17.09 and docker-compose 1.16.1.
+
+Steps
+......
+
+The first step is to copy the ``docker/dev/docker-compose.yml`` file to the root of the project. eg
```
jaywink: The file is missing from this commit, probably need to tune the `.gitignore` to exlude it from that path, but include it when in root path.

jaywink (changes requested):
Great work, thanks for this! 🎉  Some comments to move forward.

jaywink:
Also, as a generic comment. Always use a branch in your own fork and do PR's from that branch to upstream master, This makes mistakes like overwriting upstream changes less likely and also allows to work on multiple things at once 👍

ghost:
> Also, as a generic comment. Always use a branch in your own fork and do PR's from that branch to upstream master

Was using a local branch locally but good point.

Let me address the others then see what I need to do to ammend the PR

.env.example
ghost: This is a good question. Here's my reasoning:

The .env file is also used by docker-compose for its environment variables. So this means we only have one source of truth. But it's easily moved to the docker-compose file.

docker/dev/Dockerfile.django
```diff
@@ -1,12 +1,6 @@
-# We use the official Python images. This may need to get locked in, or back to Ubuntu at some point
 FROM python:3
 
-RUN apt-get update
```
jaywink: Hmm, why did you remove the package installs?

ghost: The reason for the package installs was just for XSLT and related.

Otherwise, the Docker image software should be maintained by the upsteam Docker maintainers. There are three reasons why this is the case:

1. This is the Docker data model

What I mean is that Docker images will only fetch data that's necessary to overlay on top of each other, recursively. It's more space efficient to do this

2. We're using the official Python docker images

Just as we would generally rely on upstream package maintainers in Linux or BSD to maintain upstream code, we should be leaning on the Python docker maintainers to maintain upstream docker. The only reason we pulled in additional packages was for XSLT, and as you mentioned, that's no longer necessary.

3. If the official Python image changes somehow, we're less likely to break.

It appears that Python's docker image is using Debian, but if it should change somehow, our docker images will still build properly if we aren't relying on anything other than Python itself.

That's why I removed it.

ghost:
I believe the last remaining objection was the removal of the apt-get lines from the Django image, which I've explained.

The only other changes I've made since the origial PR were to address the issues @jaywink  mentioned in his comments as well as add some additional verbosity to the Javascript watching image.

docker/dev/Dockerfile.django
jaywink: Actually it was a question:

> I'm kind of surprised libxml2-dev libxslt-dev are not needed. Those are direct dependencies for lxml to be built successfully. Maybe the Python3 image has them already?

https://github.com/jaywink/socialhome/pull/334#discussion_r145684629

At least Ubuntu requires a few extra libraries to build `lxml` (see http://lxml.de/installation.html#requirements and https://socialhome.readthedocs.io/en/latest/install_guides.html#install-system-packages (federation ie lxml dependencies)). So if the `python:3` image does work without these extra installs, it must have them already? Otherwise I can't imagine that the docker images work.

Does a clean build of this docker image without these extra installs actually work? And tests also pass?

docs/development.rst
jaywink: `docker-compose.yml` is still missing from the PR.

.env.example
```diff
@@ -11,3 +11,9 @@ SOCIALHOME_DOMAIN=127.0.0.1:8000
 SOCIALHOME_HTTPS=False
 
 MOCHA_TESTS=True
+
+# Set these variables for Docker Postgres (be sure the DATABASE_URL matches)
+POSTGRES_USER=socialhome
```
jaywink: As discussed, these variables should be imho defined in `docker-compose.yml` (which is missing btw). Otherwise we have database settings twice, and really there is no reason for developers normally to change these.

jaywink:
There are still some master commit things being reverted in this PR - please check the comments carefully, thanks.

jaywink:
@Vir-Cotto could you check the remaining comments I've made, for example using this view: https://github.com/jaywink/socialhome/pull/334/files .. all of those are still relevant. Thanks!

docs/development.rst
ghost: :heavy_check_mark:

docs/development.rst
ghost: :heavy_check_mark:  Sorry about that.

docs/development.rst
ghost: Yes, this is actually a thing to think about in production (since notifications are via mail) but it's fixed in my current version. :heavy_check_mark:

docs/development.rst
ghost: It's on the .gitignore and in the docker/dev directory. :heavy_check_mark:

.env.example
ghost: :heavy_check_mark:

docs/development.rst
ghost, line 241: Done. :heavy_check_mark:

ghost:
All the comments should now be addressed.

jaywink:
I think you maybe forgot to push your changes? I see you've checked my comments but nothing has changed regarding the comments, sorry.

docs/development.rst
```diff
@@ -203,6 +191,47 @@ Then execute the following and copy the markdown version for pasting to GitHub r
...
+``docker-compose run django manage createsuperuser``
+
+And then just
+
+``docker-compose up``
+
+Defaults
+..........
+
+The defaults are that that the Docker image will be running on port 8000 and then exposed to the host OS on the same port (ie you can browse to http;//localhost:8000 to see the Django instance running). Redis and Postfix will be running but not exposed to the host OS by default. These can be changed on the ``docker-compose.yml`` file.
```
jaywink: `Postfix` -> `PostgreSQL`

ghost:
Fixed the Postfix mistypes. I have email on the brain.

ghost:
The documentation is now fixed. I'm not sure what else needs addressing. I would recommend squashing the last few commits.

jaywink (approved):

jaywink:
Thanks! 🎉

PR #336, opened 2017-10-19
Enable to add additional third-aprty apps using .env

autogestion:


jaywink:
Hi @autogestion - thanks for the PR!

I'm not entirely sure what the use case for the `ADDITIONAL_APPS` setting would be. I would prefer not to introduce new settings unless there is a real use case. Can you give an example what it could be used for? I would imagine for adding extra functionality, in most cases one would need to fork the project and do changes anyway. Adding an app is rarely useful on it's own, in my experience, but I could be wrong.

About the `DATABASE_ENGINE`. This can already be done actually. Socialhome is configured using a `DATABASE_URL`. So you can construct the url using any database supported by Django. See https://github.com/kennethreitz/dj-database-url#url-schema .. so at least this new setting is not necessary.

autogestion:
Thanks for tip with engine!

There is an idea to create federated anti-corruption network. Previous idea, with centralized server, encountered with the problem, that it is hard to get together journalists, politicians and public organizations, they have different standards and requirements. So in federated version actors will aggregate claims on their own pods with their specific marketing strategy and share results using Diaspora protocol

To do this, there is an idea to create reusable app https://github.com/autogestion/sh_ctracker
that will allow to treat content as claims, and to use leaflet-map to show corruption with color encoding. And to make deployment (and updating sh engine) of anti-corruption pod easier, we need ability to add apps in such way

Also, there would be great to have opportunity for such app to intercept root url in some civilized way (in our case to show map on main page), maybe you can suggest some solution

jaywink:
@autogestion wow, this is highly interesting and I'm honoured you chose to use Socialhome for the base.

My first question is: why base it on Socialhome? Is it simply to get federation or is it because you want some of the other Socialhome features like content, replies, profiles, etc? If the former, it might be less work for you to implement the [federation library](https://github.com/jaywink/federation) into your application instead. This is what Socialhome uses for federation. Obviously it requires a lot of glue on the application side, which Socialhome does already.

If you still want to base it on top of Socialhome, that is fine. Just something to keep in mind that Socialhome is AGPLv3 licensed, so if you copy any code you should respect the license, for example by adding the license file to whichever files you use and indicate the origin + changes made.

Of course we could make your life as easy as possible by ensuring Socialhome can be configured in a way that your app runs without doing many modifications. If that is possible. Now I see your app install instructions require a lot of manual steps and modifying Socialhome code. I kind of get the feeling that a few settings will not be enough and you would anyway have to either maintain a fork or make those manual changes.

So final question - why not maintain a fork with the changes you need to make `sh_tracker` work? Anything that makes sense to integrate to Socialhome could then be merged in, but otherwise you could freely modify your fork to ensure `sh_tracker` works but still is able to use the features + federation of Socialhome?

autogestion:
At all, I think it will be great if Socialhome would support pluggable applications, like wordpress does. In this case it could be used for varied services, not turning in huge monster in same time. And pushing features to pluggins allows to get light-weight server https://github.com/jaywink/socialhome/issues/340

We decided to take engine as a base, to focus on mapping issues, without need to develop and support additional features, which may be useful to users, such as blogging, discussions etc. As you mentioned before "Obviously it requires a lot of glue on the application side, which Socialhome does already", that's why)

Socialhome was chosen because it uses same stack (Django, DRF, Postgres) as centralized server version (library was implemented on the basis of copypast), and because it is alive

So, if making fork, we would have to support all it's codebase, and this is not what we need)

jaywink:
I actually did not think about a pluggable apps use case at all, but hey why not, it's a great idea. The only problem I think is that Socialhome is in flux still. Nothing is guaranteed to be absolutely stable until we hit 1.0 version - even API or internal models. Thus, creating any kind of way to have pluggable apps is going to be tricky.

I think we can move with this at least partially to what you need. We can solve things case by case as they come up. I'll make a few comments on the setting itself, then we can merge this and look at other ways to help the pluggable apps use case 👍 

> Also, there would be great to have opportunity for such app to intercept root url in some civilized way (in our case to show map on main page), maybe you can suggest some solution

I suggest a setting `SOCIALHOME_HOME_VIEW` that is default to `None` and allows giving a full Python import string for a custom view, for example `"myapp.views.AwesomeHomeView"`.

Then we check this setting [here](https://github.com/jaywink/socialhome/blob/f048f0adc5c07ce48e562351ae78dabf799a2622/socialhome/views.py#L18) and if it is set, import the view and render it instead of the current home view, just like we already do there for other views. The reality is SH doesn't really have a home page - it depends on user and their settings.

.env.example
```diff
@@ -11,3 +11,5 @@ SOCIALHOME_DOMAIN=127.0.0.1:8000
 SOCIALHOME_HTTPS=False
 
 MOCHA_TESTS=True
+
+ADDITIONAL_APS=
```
jaywink: No need to have this in the example file since it defaults to none (and setting it is not a common use case).

config/settings/common.py
```diff
@@ -467,3 +467,7 @@
         ("large", "crop__300x300"),
     ],
 }
+
+ADDITIONAL_APS = env("ADDITIONAL_APS", default=None)
```
jaywink: I would prefer Socialhome settings prefixed with `SOCIALHOME` to avoid conflict with third-party apps and, well, to be more clear what their scope is. So perhaps `SOCIALHOME_ADDITIONAL_APPS`.

Also, please place the setting where the other `SOCIALHOME` prefixed settings are in `common.py` 👍 

Also, please add documentation for the setting [here](https://socialhome.readthedocs.io/en/latest/running.html#configuration), in the correct place alphabetically. Explanation should highlight this is optional to set.

jaywink (changes requested):

autogestion:
Last commit is for including urls. Looks like this all that we need, to provide an opportunity for pod admin to keep sh instance updated, and use pluggable apps same time, without need to maintain fork :)

docs/running.rst
```diff
@@ -211,6 +211,28 @@ Default: ``socialhome.local``
 
 Must be set to your Socialhome instance domain. Used for example to generate outbound links.
 
+SOCIALHOME_ADDITIONAL_APPS
```
jaywink: I hate to be picky but it's really important to keep the settings in alphabetical order. We're going to have many of them, and it will be annoying to scroll if they're not in good order. Order is more important IMHO than grouping. So SOCIALHOME_DOMAIN after SOCIALHOME_ADDITIONAL_APPS_URLS and SOCIALHOME_HOME_VIEW after SOCIALHOME_HTTPS 👍

docs/running.rst
```diff
@@ -211,6 +211,28 @@ Default: ``socialhome.local``
...
+
+Allows to use additional third-party app url-conf, string with two comma-separated values, url prefix and path to urlpatterns, for example ``myapp/,myapp.urls``.
+If you need to include urls from more than one app, this could be done by creating intermediary app which aggregates urls
+
+SOCIALHOME_HOME_VIEW
+.....................
+
+Default: ``None``
+
+Allows to use on main page custom view from third-party app, string with path to view, for example ``myapp.views.AwesomeHomeView``
```
jaywink: Nitpick missing period at the end of the sentence.

docs/running.rst
```diff
@@ -211,6 +211,28 @@ Default: ``socialhome.local``
...
+
+Allows to plug in additional third-party apps, string with comma-separated values, for example ``django.contrib.gis,myapp``.
+
+SOCIALHOME_ADDITIONAL_APPS_URLS
+.....................
+
+Default: ``None``
+
+Allows to use additional third-party app url-conf, string with two comma-separated values, url prefix and path to urlpatterns, for example ``myapp/,myapp.urls``.
+If you need to include urls from more than one app, this could be done by creating intermediary app which aggregates urls
```
jaywink: Nitpick missing period at the end of the sentence.

socialhome/views.py
```diff
@@ -15,6 +17,11 @@ class HomeView(TemplateView):
 
     def get(self, request, *args, **kwargs):
         """Redirect to user detail view if root page is a profile or if the user is logged in"""
+
+        if settings.SOCIALHOME_HOME_VIEW:
```
jaywink, line 21: Just want to check that you feel it's better that we don't respect the user setting for landing page if the custom home view is set? Is this what was intended?

Also, it might be better to not render the setting in the case that custom home view is set, otherwise I think it would confuse users since their setting would have no effect at all. Unfortunately, that is tricky since we don't have a proper custom form for that, it comes from `django-dynamic-preferences` app. Would probably require overriding the form and then excluding the field there in init or something.

Something to think about. I'm fine with leaving this detail to later, when users complain :) But maybe think whether you intended to override the user landing page totally or just redirect for non-authenticated users to the custom home view.

jaywink (changes requested):
3 very minor issue, 1 question - then good to merge!

docs/running.rst
jaywink: Wait, ignore the latter, O is before T of course 😆 But the first part is relevant, A before D.

socialhome/views.py
autogestion: Our module was convinced as corruption monitor, user opens landing page and see that this city became red (corruption here increasing) , and that district became green (good news). You can play with it on our test server http://test.acts.pp.ua:8005/  (sometimes it unavailable)

So this custom view will be in use by both registered and unregistered users (I think any module of such kind will need focus on custom view). In current sh implementation, respecting user setting will make custom view unavailable for registered users at all, except by logging off.

The best way is to find some solution, that dynamic-preferences will support custom view (and pod admin will have ability to change default view for user). As I see, it could be a little tricky.
Eventually, project is still in development, maybe in future some cool solution will become obvious, and we can leave it as is)

jaywink (approved):
Awesome, good to go! 👍 🎉 

Let's expand where necessary. As long as these expansion points don't make the core unnecessarily complex, they're welcome to fit in a larger ecosystem of platforms.

Thanks!

autogestion:
fixed)

jaywink:
Awesome, thanks! 🎉

PR #337, opened 2017-10-19
Reduce layer count and remove inline comments

0rzech:
Few changes reducing layer count, removing deprecated inline comments etc. to adhere a bit more to [best practices for writing Dockerfiles](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/). From Docker 17.05 onwards, using [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) should be a better way to minimize unnecessary image content count.

### DIFF
--- a/docker/daphne/Dockerfile
+++ b/docker/daphne/Dockerfile
@@ -1,35 +1,42 @@
 FROM ubuntu:16.04
 
 # Generate some locales
-RUN apt-get update && apt-get install -y locales && rm -rf /var/lib/apt/lists/* \
-    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+RUN set -eu && \
+    apt-get update && \
+    apt-get install -y locales && \
+    rm -rf /var/lib/apt/lists/* && \
+    localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+
 ENV LANG en_US.utf8
 
-# Install OS packages
-RUN apt-get update && apt-get install -y \
-    # Direct dependencies for socialhome
-    git python-virtualenv python3-setuptools python3-dev build-essential libpq-dev \
-    # Federation
-    libxml2-dev libxslt-dev lib32z1-dev \
-    # Redis TODO move to own container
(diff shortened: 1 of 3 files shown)


docker/daphne/Dockerfile
```diff
@@ -1,36 +1,43 @@
...
-    # Direct dependencies for socialhome
-    git python-virtualenv python3-setuptools python3-dev build-essential libpq-dev \
-    # Federation
-    libxml2-dev libxslt-dev lib32z1-dev \
-    # Redis TODO move to own container
-    redis-server \
-    # Generic stuff for Ubuntu
-    wget apt-transport-https apt-utils
+# Install OS packages:
+# - direct dependencies for Socialhome
```
jaywink, line 13: Having comments this way kinda makes them not so useful since the relation of the comment and what is being commented on is being lost. I would prefer keeping the comments before the line itself unless there is a technical reason that it doesn't work. The comment could postfix the line too if before doesn't work in some situation?

jaywink (changes requested):
Thanks! Just one comment.

Note that the `dockerize` branch is currently a bit stale and it is unsure whether these changes will end up as the production images. Once the dev images end up in master in #334 , these can be continued on. Help is welcome :) Might be a good idea to pop into the chat to discuss things if there are uncertainties.

These changes definitely improve the dockerize branch and I'll gladly merge them in once the one comment is checked through 👍

docker/daphne/Dockerfile
0rzech: When I build image with these inline comments, I get notification from docker (possibly newer version than yours) that this is deprecated and will stop working in future versions. This is the only reason I've moved them up.

jaywink:
OK fair enough. Can't see your comment here but got a mail notification. If docker is really deprecating this style of comments, let's fix it as you suggest.

I'll merge this into the dockerize branch. Let's continue there 👍 Help welcome with finalizing the docker images for production. Please note however the work ongoing in #334 . Even if the images are different (and should be), would be nice to store the files in a similar fashion without conflicting and having them in two places.

Thanks!

jaywink (approved):

PR #346, opened 2017-10-23
[WIP] Infinite scroll on streams

christophehenry:


socialhome/streams/app/components/Stream.vue
```diff
@@ -82,19 +94,30 @@ export default Vue.component("stream", {
...
+            } else if (this.$store.state.streamName.match(/^profile_all/)) {
+                options.params.id = this.currentBrowsingProfileId
+                this.$store.dispatch(streamStoreOperations.getProfileAll, options)
+            } else if (this.$store.state.streamName.match(/^profile_pinned/)) {
+                options.params.id = this.currentBrowsingProfileId
+                this.$store.dispatch(streamStoreOperations.getProfilePinned, options)
+            }
+        },
+        loadMore() {
+            this.loadStream(this.$store.state.contentIds[this.$store.state.contentIds.length - 1])
```
jaywink: This looks wrong. It should the *smallest* `throughIds`, not the last element in `contentIds`. You can use something like [this](https://github.com/jaywink/socialhome/blob/7d3b0bd0b93e5580706686a38e6d3987767e0d2a/socialhome/static/js/streams.js#L116).

Last element doesn't mean smallest. It might in practise, but it's not a good idea to rely on the order. We will pop stuff into the list on top of the stream, for example.

jaywink (changes requested):
Thanks for the PR! However, I'm not sure this works as it should. When does the load more call trigger?

A reminder of the requirements (= how it works currently, which is perfect):

* the load more triggers on the 5th last content grid element in the stream once it becomes visible.

To me it seems now it requires scrolling to the end of the page, which is not what we want, as explained on IRC.

socialhome/streams/app/components/Stream.vue
christophehenry: Ah, yes, indeed.

jaywink:
But the element it triggers on should be a `grid` element, the 5th last. So AFAICT it's not in the right place. I'm thinking some simple check against state when looping over grid elements and if rendering the 5th last one - add the load more directive. That should move it forward when new elements are added, in theory :)

Btw, please if possible, when opening PR's that are not ready to merge in your opinion, prefix the PR name with "[WIP]" for example - also good to list TODO's in the description so I know what is missing. Thanks!

christophehenry:
Ok, so we have a problem: it seems Vue can't dynamically bind directives like `v-infinite-scroll` on elements. When rendered, directives bound with `v-bind` are treated like other attributes which means infinite loading won't be triggered. My current solution is to add the directive on every element, but do something only if it was called on the 5th last element. It's not nice and we need to see with usage if it doesn't slower the page. There is a deboucing to prevent the function to be called more than once every second. So it should not be too heavy.

To prevent the stream to load more content when there's nothing more to load, I added a `loadMore` variable to the store that is set to false if the last fetch returned an empty array.

christophehenry:
Tests OK. Should be ready for review.

jaywink:
Checked this locally and it seems to be generating an awful lot of executed JS. `methods` are not cached, so scrolling is basically executing a lot of code all the time. You can see it by opening the console and seeing the duplicated console.log outputs come in hundreds of times while scrolling down.

I think we need to think of some other way :(

How about calculating a `loadMore` bool on the `content` object itself when loading the items in `streamStore.onSuccess`? This way the component can just look at the attributes of the content and know whether it needs to have a load more trigger on it.

Then when executing the load more method on the component (scrolling reaches right point), we clear the attribute on the content by dispatching an action.

One other thing that comes to mind is not attaching the load more trigger to `StreamElement` but palcing a `LoadMore` (invisible) component just before the last fifth `StreamElement`, and then destroying it after load more triggers.

I would not mind trying these routes unless you have a better idea or you think it is not going to work. Right now unfortunately the performance is worse than the jQuery streams when scrolling the stream.

I also see there is a lot of overlap with #357 which got worse due to the refactoring I had to make. Let's try merge this first, but if not possible, I'll gladly do the conflict fixing 👍

christophehenry:
This solution does not have performance penalties on my computer. But that's what I feared: my computer is a beast :/
We need to think about phone usage first. If this solution is not viable, then I need to rework it. I'd like to try one last thing, though. I didn't optimally used the [`infinite-scroll-disabled`](https://github.com/jaywink/socialhome/pull/346/files#diff-b5ee95312b7126cfe16e58eaf5c7c384R51) property. [This condition](https://github.com/jaywink/socialhome/pull/346/files#diff-b5ee95312b7126cfe16e58eaf5c7c384R73) should be taken in account by it as well as [this one](https://github.com/jaywink/socialhome/pull/346/files#diff-f6bf769c688863e153cb8f8d589a3803R87) (is it what you are proposing?).

> One other thing that comes to mind is not attaching the load more trigger to `StreamElement` but palcing a `LoadMore` (invisible) component just before the last fifth `StreamElement`, and then destroying it after load more triggers.

I don't like this solution. It's not the Vue way. It's a very jQuery-way of thinking.

jaywink:
>  I'd like to try one last thing, though. I didn't optimally used the infinite-scroll-disabled property. This condition should be taken in account by it as well as this one (is it what you are proposing?).

I think though that it's not possible (performance) to make the stream listen to load more on every single element. It will just start to die when there are hundreds of elements on the screen. Somehow we must ensure there is only one element at a time that is being tracked.

> It's not the Vue way. It's a very jQuery-way of thinking.

Why is it a jQuery way of thinking? It's the components reacting to changes in the store. But, anyway, I think the first option I mentioned would be better, if it works, to make only one element at a time have the load more trigger attached to it.

christophehenry:
> It will just start to die when there are hundreds of elements on the screen.

I'm not that sure. The plugin has a throttle delay. Which means some JS code is executed whenever you scroll anyway. And Vue is thought to be a very performant framework. For instance, the interface is rendered periodically (like a video game).

I think the bes is to test. Do you think you can write me a function to generate, say 10,000 posts?

> Why is it a jQuery way of thinking?

Manually attach and remove elements at some place in the DOM. Vue is a declarative framework. You should write your template according to your data, not think yourself about how elements should be rendered.

If this solution is not viable anyway, I think the best is to let go the idea of attaching the event to the fifth last post and attach it to the loader like I originally did. I can still increase the distance so that the rest of the page will always be loaded before the user reaches the end of the page.

jaywink:
> I'm not that sure. The plugin has a throttle delay

Yes but having *anything* execute 100 times on a page every time someone scrolls versus having it execute 1 time is what my point is about. We only need one load more trigger, so we should find a way to do it. The best solutions are never the easiest :)

> I think the bes is to test. Do you think you can write me a function to generate, say 10,000 posts?

```
from socialhome.content.tests.factories import PublicContentFactory
for x in range(10000):
    PublicContentFactory()
```

Something like that off the top of my head :) If you want longer content you can specify some random stuff in `text` attribute for example, for each iteration.

> Manually attach and remove elements at some place in the DOM. Vue is a declarative framework. You should write your template according to your data, not think yourself about how elements should be rendered.

What I meant was putting the load more trigger behind a div that has a` v-if` checking for `content.load_more` attribute, which would be pre-calculated to be `true` only for the last 5th element by the store. IMHO it's better to offload calculations to one place rather than calculate something in each iteration of rendering. Or even worse, each iteration of scrolling.

So, when the `content.load_more` attribute turns from `true` to `false`, the `v-if` re-renders the element, removing the load more. I think my explanation was a bit bad. And maybe it wouldn't work, I don't see why not though, unless the infinite scroll plugin has some weird stuff.

> If this solution is not viable anyway, I think the best is to let go the idea of attaching the event to the fifth last post and attach it to the loader like I originally did. I can still increase the distance so that the rest of the page will always be loaded before the user reaches the end of the page.

Unfortunately we can't go with that. It would be a worse experience than currently. And we only want to make things better :) As said before, a really long say 2000px content would basically force the user to scroll down to the end of that content and then scroll back up again to read what appeared. Having the load more the 5th last guarantees in a 4 column layout that you always trigger it before you see the end.

Don't worry, there will be a way. Surely jQuery can't win here? ;)

christophehenry:
> What I meant was putting the load more trigger behind a div that has a `v-if` checking for `content.load_more` attribute, which would be pre-calculated to be `true` only for the last 5th element by the store.

How come I didn't think about that!? >.<

christophehenry:
Ok, so I dropped `vue-infinite-scroll` in favor of `mugen-scroll`. I hope it works better.

jaywink:
Will start checking if I could help with this now 👍

socialhome/streams/app/components/Stream.vue
```diff
@@ -76,25 +77,43 @@ export default Vue.component("stream", {
         },
     },
     methods: {
+        loadMore() {
+            let contentId = this.$store.state.contentIds.slice(-1)[0]
```
christophehenry, line 81: `this.$store.state.contentIds[this.$store.state.contentIds.length - 1]`

jaywink:
Firstly, sorry this took so long for me to resolve. This branch is not compatible directly with master any more so it was tricky to test it reliably. I ended up taking my "load more" branch and implementing both versions of the load more trigger there. I've spent a few hours mucking around trying to see which one is better and ensuring both do the same thing.

Unfortunately I'm not really sure how I could get reliable metrics out, TBH, most of the weight in the stream is caused by the constant layouting work by Masonry. This starts to become insanely heavy once the stream has several hundred items. From my testing it feels like the stream grinds to a quicker halt with the solution in this PR, using the `isFifthLast` check. Having now worked with both solutions I'm not entirely sure why. It could be the deep store lookup for each content during every re-render of every item in the stream. While my solution only has to check a boolean in local state for every re-render of every item. So a difference of running the following hundreds/thousands of times every time a load more happens:

    this.$store.state.contentIds[this.$store.state.contentIds.length - 5] === this.contentId

vs

    this.content.hasLoadMore

Precalculating probably helps a bit in this case. It's better to take as much as possible off the components rendering as we can and instead precalculate stuff either in the backend or when store state mutates, IMHO.

I made sure the rest of the code is pretty much the same while testing, only changed the way the infinite scrolling div checks if it should render and of course where in the state we store the load more flag (root vs content). This was possible by refactoring the stream load dispatching itself into an action, thus getting rid of the need to emit an event like done in this PR.

I also of course had to pick up the stream loading code from this PR, which I made a commit using your name, since I took it from here. Please let me know if this is not ok.

So tl/dr, we're going to use the `content.hasLoadMore` precalculation way as in the other PR since I have a strong gut feeling plus some not so scientific testing that it is less heavy on the rendering when the stream is hundreds of items in size. But otherwise, I took everything I could from this PR in regards to actually fetching the new content.

Since this PR is conflicting quite heavily, I will close it in favour of #362 , sorry :/ I really appreciate your work on the Vue streams and I'm truly sorry we managed to come up with two different solutions here, and that other code changes caused this branch to conflict quite badly.

I will still pick up some stuff from here and make commits in your name to the other PR, if ok with you. Some of the tests you wrote here can be picked in the other PR, so writing them was not wasted effort.

christophehenry:
Ok.

christophehenry:
I'm saving this branch into [`vuejs-infinite-scroll-bckp`](https://github.com/christophehenry/socialhome/tree/vuejs-infinite-scroll-bckp) if you need the code.

Issue #339, opened 2017-10-20
Change the envelope icon for replies in streams


Vir-Cotto:
In the streams UI, we use an envelope to represent replies, which are comments.

Envelopes usually indicate email, or sometimes notifications. We're already using FontAwesome- why not switch to the comment icon http://fontawesome.io/icon/comments/

jaywink:
Sounds an ok change to me. What do you think @christophehenry, you've been interested in design tweaks?

christophehenry:
I agree with @Vir-Cotto.

jaywink:
OK, let's change it then! @Vir-Cotto feel free to make a PR 👍 Note that there are several places (current stream, vue stream), searching through with the old class is the best way to find them.

jaywink:
Closed via #347

PR #347, opened 2017-10-23
Comment icon (#339)

ghost:
I believe this should solve #339

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -45,6 +45,10 @@ Changed
 
   This makes edits federate more reliably to some remote platforms that support edits.
 
+* Replaced fa-envelope with fa-comments on Replies
+
+  Replies are what many systems call "comments" so using the corresponding icon.
+  
 Fixed
 .....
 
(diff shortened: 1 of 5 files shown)


ghost:
Due to some issues on my end, this has an accidental dependency on the docker-dev image PR. Sorry about that.

jaywink:
Basically looks good (though a screenshot would be nice when touching visual stuff - I think this time we'll survive without one :)), but we need to get rid of the other commits. Maybe like this:

    git remote add upstream https://github.com/jaywink/socialhome.git
    git fetch upstream
    git checkout upstream/master
    git cherry-pick f9de743b74f12a785c5f53e27fe5636695ed2e77
    git branch -D comment-icon
    git checkout -b comment-icon
    git push origin --force
    
Might be a shorter way but this at least should do it :)

Also, could you add a [changelog entry](https://socialhome.readthedocs.io/en/latest/changelog.html#changed) to the `docs/changelog.rst` unreleased changed section? Just a simple "reply icon has been changed etc" and then link to the issue that is being solved (#334).

Also, please add yourself to `CONTRIBUTORS.txt` as this will be your first merged in code contribution. The file is not alphabetically ordered.

Thanks!

jaywink:
Thanks! 🍻

Issue #327, opened 2017-10-14
Remote user profile save should fix image urls


jaywink:
All our `Profile` image urls should be absolute. Some remote delivered profile urls are relative. We should fix them, for example the following:

    /assets/foo.bar

becomes

    https://domain.tld/asserts/foo.bar

The domain.tld comes from the remote profile handle.

amirkarimi:
I would like to work on this.

jaywink:
@amirkarimi Awesome!

Yes, you got it right. Basically the lines are [here](https://github.com/jaywink/socialhome/blob/9a83abcfb9cb43139f5fb86a437926358730536e/socialhome/users/models.py#L281-L283):

    "image_url_large": safe_text(remote_profile.image_urls["large"]),
    "image_url_medium": safe_text(remote_profile.image_urls["medium"]),
    "image_url_small": safe_text(remote_profile.image_urls["small"]),

If the `remote_profile.image_urls` urls contain a relative url, it needs to be made absolute using the domain of `remote_profile.handle`. We can safely assume https is the protocol for all remote urls.

amirkarimi:
Sorry for the delay. I started working on this issue but realized that it's been implemented [here](https://github.com/jaywink/socialhome/blob/master/socialhome/users/models.py#L210). `safer_image_url` will make relative the URLs absolute so as long as `safer_image_url_(small|medium|large) is used the URLs would be OK. In another world, it's been made absolute on each attribute access not on load/save.

The question is; providing the above information, is this issue still valid? If so, to close this issue, should we:
- Make URLs absolute at load time [here](https://github.com/jaywink/socialhome/blob/9a83abcfb9cb43139f5fb86a437926358730536e/socialhome/users/models.py#L281-L283)
- Remove `safer_image_url_*` and replace all usages with `image_url_*`

jaywink:
Yes you are completely right. Sorry for not mentioning that in the issue, I was obviously aware of the `safer_` wrappers but forgot to mention them in the issue :(

The aim is to ensure `image_url_*` fields always contain an absolute URL, so we don't have to wrap them in any way. Right now this is a problem in the REST API since we don't use the wrappers there at all (it would add more complexity to each serialization).

To close this issue it's necessary to convert the remote profile url to an absolute one, as indicated in your first option. You can re-use the "safer_" code if you want 👍 Then we can remove the "safer_" properties and just use the image url fields directly in the UI. That is kind of optional though, not directly related to this issue.

[...]

PR #348, opened 2017-10-28
Make profile image url absolute

amirkarimi:
https://github.com/jaywink/socialhome/issues/327

### DIFF
--- a/socialhome/users/models.py
+++ b/socialhome/users/models.py
@@ -272,12 +272,13 @@ class Profile(TimeStampedModel):
         return ""
 
     @staticmethod
-    def absolute_remote_url(remote_profile, image_name):
-        url = safe_text(remote_profile.image_urls[image_name])
+    def absolute_image_url(profile, image_name):
+        """Returns absolute version of image URL of given size if they wasn't absolute"""
+        url = safe_text(profile.image_urls[image_name])
 
         if url.startswith("/"):
             return "https://%s%s" % (
-                remote_profile.handle.split("@")[1], url,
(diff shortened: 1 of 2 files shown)


socialhome/users/models.py
```diff
@@ -272,15 +272,25 @@ def get_last_name(self):
         return ""
 
     @staticmethod
+    def absolute_remote_url(remote_profile, image_name):
```
jaywink: Could the signature be `absolute_image_url(profile, image_name)`? Reason: while we're using this for remote profile parsing, the functionality in the method doesn't actually have anything strictly to do with remote profiles. But it does have everything to do with image url's, so calling it `absolute_image_url` would make it easier to see right from the name what it does. With the current name my first thought is that this returns "an absolute url for the remote profile itself".

Also, a short docstring would be nice, just to indicate what it does (= returns absolute version of image ulr of given size).

jaywink (changes requested):
Thanks for working on this! One change request, then good to go!

jaywink (approved):

jaywink:
Perfect! Thanks for this <3 Did you also want to work on removing the `safer_` method usages (requires also doing a data migration for any profile image urls that are relative) or shall I log it as a follow-up issue?

amirkarimi:
@jaywink I would like to work on it but as I might be a little bit busy, I think it's better to create another issue. I'll pick it up ASAP.

jaywink:
@amirkarimi logged #374 , sorry forgot about this for a while. If you're interested in tackling it at some point. Thanks again for your contribution!

Issue #314, opened 2017-10-04
Federate out edited timestamp when editing content


jaywink:
Currently when federating content edits, the created timestamp is sent out. Send instead the edited timestamp.

This makes edit federate more reliably with for example Hubzilla.

Vir-Cotto:
This seems to be passed to the Federation library as created_at. This may be a Federation question, but I don't see a modified_at option in Federation, so should this always be the modified timestamp (if there is one)?

Second question- we could fix this just in make_federatable_content but we're using a calculated property on the objects called effective_creted. Should this be fixed in the federation part or in the model?

jaywink:
>  This may be a Federation question, but I don't see a modified_at option in Federation, so should this always be the modified timestamp (if there is one)?

Currently there is no `modified_at` or similar in the federation library because the [Diaspora protocol](https://diaspora.github.io/diaspora_federation/entities/status_message.html) doesn't have one. Since [ActivityPub does](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-updated), we will likely at it when AP support is added to the federation library.

For now, just passing our modified timestamp to the federation entity `created_at` is the goal to do. This is how we transmit an edit by just sending the object again, since the Diaspora protocol doesn't have edit. This works between SH instances and to Friendica and Hubzilla, but not Diaspora which just rejects the re-sends.

> Second question- we could fix this just in make_federatable_content but we're using a calculated property on the objects called effective_creted. Should this be fixed in the federation part or in the model?

Good point, I forgot we're not using the `created` directly. It seems `effective_created` does not have other usages. I can't remember the history why we want to transfer out a `remote_created` if it exists, since in the UI and everywhere we use `modified` anyway.

I suggest lets just replace `effective_created` with `modified` when creating the federation payloads and remove the `effective_created` property from the `Content` model as unnecessary to have.

The `Content.remote_created` timestamp collection still makes sense since sometimes we might be interested in the remote timestamp too. Sometimes a server might have content created at a time but federation only happens later, so these could differ.

Does this clarify?

jaywink:
Actually, how about we just change 

```
@property
    def effective_created(self):
        if self.remote_created:
            return self.remote_created
        return self.created
```

To:

```
@property
    def effective_created(self):
        if self.remote_created:
            return self.remote_created
        return self.modified
```

This would solve the problem entirely?

Vir-Cotto:
[...]

PR #355, opened 2017-11-01
Replace effective_created with modified in the _make functions

ghost:
This solves #314 I believe


jaywink (changes requested):
While this solution could do also, did you see [my comment](https://github.com/jaywink/socialhome/issues/314#issuecomment-340764545) yesterday for an even better solution? I think it would be better to do it that way since it supports remote entities better (= they can't be edited locally).

Could you adjust to that comment? Thanks

PR #357, opened 2017-11-01
Add replies to vue.js streams + other improvements

jaywink:
* Add replies to Vue.js streams
  * Organize replies first by original content then shares
* Add shares endpoint to Content API
* Add share replies to Content API replies endpoint
* Add share_of to Content API serialization
* Highlight reaction icons with active colour in vue.js streams
* Remove unnecessary select_related in contentqueryset

TODO:
* [x] Add moar tests for new and changed functionality
* [x] Don't show reply to share button if the share has no replies
* [x] Add spinner when loading the replies

Refs: #202, #206


jaywink:
@christophehenry if you would like to review this PR as it's related to the vue.js streams then please feel free to do so 🎉

christophehenry (changes requested):
I prefer having a separate `replyStore` rather than putting replies in the `streamStore`. Let's keep things simple and respect separation of concerns :wink:

jaywink:
> I prefer having a separate replyStore rather than putting replies in the streamStore. Let's keep things simple and respect separation of concerns 😉

I did consider that. There are a few reasons I think are valid that I didn't go this way:

* We then also should have a `shareStore` and in the future `reactionStore` and we should extract `profileStore` at some point.
* Replies *are* an as important part of the streams as is the content itself. For example, replies do not exist outside streams, they are only in streams. If we separate stores by object in the streams app, we should then also rename `streamStore` to `contentStore`, since streams isn't an object at all and actually there is no data for streams.
* Content and Replies *are* the same model. This is why `replies` are currently stored in `contents` in this PR. This is how they are stored in the backend as well, they have exactly the same data schema. Of course, you could say the frontend design should not care about the backend data model. But in this case I don't see why we should make things more complex if logically the data is actually the same.
* We re-use `StreamElement` for both content and replies (as we do in the current stable code too). Having the store different would either mean duplicating the component or having extra `if/else` in the component.
* In the future we will (hopefully) replace the whole application frontend with vue.js. This means we will for sure have to refactor the store into better logical parts. However, vuex defines the following about state:

  > Vuex uses a single state tree - that is, this single object contains all your application level state and serves as the "single source of truth". This also means usually you will have only one store for each application. A single state tree makes it straightforward to locate a specific piece of state, and allows us to easily take snapshots of the current app state for debugging purposes.

  https://vuex.vuejs.org/en/state.html

  Organizing the single store (right now we actually have two) should happen through modules according to Vuex: https://vuex.vuejs.org/en/modules.html 

I think it's important that we started first with the streams to kind of prototype the Vue frontend to prepare for the full rewrite. When that happens, we should anyway refactor the store so that it fits into the "single store with modules" concept that Vuex highlights in their docs as the way to go. So, adding new stores at this point doesn't feel like a good thing to do. Of course I *could* add a module to the current stream store, but that would be confusing, since `content` and `replies` *are* the same models.

Thanks for the review btw, would love to have more reviewing - makes me need to explain things, since I've often made poor choices here due to no cross-checking ;)

christophehenry:
I get your point about the models. I got that was the reason when reviewing first. But even though models are the same, these are conceptually different objects. And you treat them as conceptually different objects in your code which leads to inefficient code (the getters for instance or the `if`s to determine whether to display content on the stream) and you fetch them differently. I'm concerned it will considerably slow down the stream when we reach a certain amount of content. It is especially critical on mobile as we won't use distinct pages. In addition, the code is longer and harder to understand. Let's stay pythonic on this: concise and easy to understand code.

As for the components, as long as you keep the same state structure, nothing prevents you to reuse components with a different store.

On the conceptual side, I'm not fond of that "single source of truth" thing. First, it makes the model more complex and harder to reason with. I find that assembling a bunch of modules like that produces spaghetti code.

christophehenry:
You know what? After all, that single-source-of-truth thing is not a bad idea. It makes the application to run fully off-line without problems. It's actually quite clever.

socialhome/streams/app/stores/streamStore.js
```diff
@@ -13,7 +13,23 @@ Vue.use(Vuex)
 function onSuccess(state, payload) {
```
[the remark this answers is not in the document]
jaywink: Actually, thinking about this, as I was starting to split it up. Shouldn't we not be using an `onSuccess` at all? AFAICT we should be putting all this code into our mutations. What do you think? I know this was a first version, just thinking if it's correct to keep as we move forward.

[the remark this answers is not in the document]
jaywink: OK, ignore that last comment. Did some playing around and I think the solution you've provided before is the most optimum one.

Also, the feedback about `contents` not having everything was good. It is cleaner to have them separate, so I did some refactoring. To avoid code duplication I had to go a bit deeper, touching quite a few components in the process. Only broke 27 tests :D Will fix tomorrow and hopefully get this finally merged. Re-review welcome.

socialhome/streams/app/tests/fixtures/jsonContext.fixtures.js
```diff
@@ -73,4 +75,4 @@ const getContext = function (args = {}) {
     })
 }
 
-export {getFakeAuthor, getFakePost, getProfile, getContext}
+export {getFakeAuthor, getFakeContent, getProfile, getContext}
```
christophehenry, line 80: At some point, we'll need to generate fixtures from Django models directly. It avoids false-positives when the JSON changes.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -53,7 +53,7 @@ import Vue from "vue"
 
 export default Vue.component("author-bar", {
     props: {
-        contentId: {type: Number, required: true},
+        content: {type: Object, required: true},
```
christophehenry, line 56: Why not use the store directly?

jaywink: Because whatever is passed to `StreamElement` and the below components down the line could be either in `state.contents` or `state.replies`. Since these were separated to make the store cleaner, it means the components must receive the object directly to avoid if/else in the properties (and passing an extra type to the component).

christophehenry: Can you then specify each property so that we benefit from type checking? If you need to modify properties, you can use the [`.sync` modifier](https://vuejs.org/v2/guide/components.html#sync-Modifier) on `v-bind`.

jaywink: I'll have a look at that, good point. Though only required ones should be given, otherwise we break the UI every time the `Content` serializer changes - which is every week currently ;)

jaywink: Seems to me it's not possible to define a validator for `object` requiring only certain keys to be present. One possibility is to write a [validator function](https://vuejs.org/v2/guide/components.html#Prop-Validation). I'm not honestly sure it's worth the effort since the function would be unique for each component. Some of the components require different properties available on the object.

christophehenry: If the object will be that different whether it's a post or a reply, then maybe it is worth writing two seperate templates. Otherwise, just detail all the properties and bind the whole object using `v-bind` ;)

jaywink: The whole point of reusing the `StreamElement` component is to ensure we don't *have to* write separate components that will be 98% the same lines of code.

Can you give an example of `v-bind` in this case? AFAICT we are already doing it by passing in the content with the `v-bind` shorthand `:content` from the parent. But I might be [misunderstanding the docs](https://vuejs.org/v2/guide/components.html#Dynamic-Props) :)

jaywink:
Tests - check! I'll add the spinner and squash the other todo tomorrow. Can then hopefully merge this monster and help with the load more 👍

socialhome/streams/app/components/AuthorBar.vue
christophehenry:
```js
props: {
  author: {type: Object, required: true},
  user_is_author: {type: Boolean, required: true}
  user_is_following_author: {type: Boolean, required: true}
  // ...
}
```
and
```js
<author-bar v-bind="content"/>
```
should do.

christophehenry: Also, at some point, I'd like we speak again about snake case in the API. Makes the whole front-end inconsistent and it's not common. Most HTTP APIs I've seen so far always used camel case. There must be some option to make Django REST Framework deliver camel case. What about [Django REST Framework JSON CamelCase](https://github.com/vbabiy/djangorestframework-camel-case)? It's [mentionned in the documentation](http://www.django-rest-framework.org/api-guide/renderers/#camelcase-json). So, can't be that bad.

jaywink: Interesting, have to try this tonight at home. Basically, we're already using `v-bind` as follows (`:content="content"` is the same as `v-bind="content"` according to the docs.. I mean, why have one way of writing something when you can have many!).

So how does the current code even work? Looking at your example [this should crash](https://github.com/jaywink/socialhome/blob/vue-replies/socialhome/streams/app/components/AuthorBar.vue#L56), but it works fine. I'll see what happens if I extract the object properties.

I'm concerned however that doing it that way will create more mess and less clarity. For example, if we reference `id` in `AuthorBar`, how can the reader know it's Content ID. I would myself assume Author ID just by looking at the component. Passing in `contentId` as before or `content` as in the new code makes it clear what is being referenced. Having `id` as a property is vague.

Regarding the camelCase in the REST API. I'm afraid this is a no. The reasons are the same:

* There is [no standard](https://stackoverflow.com/a/25368854/1489738) in [JSON](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) that objects should have snake_case or camelCase attributes.
* Similarly there is no standard for REST either (well, it's not really a data format specification...)
* The Socialhome API is not just for the frontend. And not just for JavaScript consumers.
* Adding camelCase to REST framework is only possible via a third-party dependency, as you said. It is not justified to add new dependencies just because the JS code "looks bad". The extra dependency could some day stop us from upgrading to a newer DRF version, for example, or could [contain bugs](https://github.com/vbabiy/djangorestframework-camel-case/issues?utf8=%E2%9C%93&q=is%3Aissue).

Don't worry, the snakes wont bite ;)

christophehenry: > Basically, we're already using `v-bind` as follows (`:content="content"` is the same as `v-bind="content"` according to the docs.

No, `:content="content"` is a shorthand for `v-bind:content="content"`. `v-bind="content"` binds `content` properties to matching component properties.

> Don't worry, the snakes wont bite ;)
 
No, but in this case, it forces front-end to write some properties in snake case and other in camel case :/

jaywink: Yeah the `v-bind="content"` works 👍 Thanks!

Now, the only problem is, I'm really not sure this is a good way to go. What is the actual benefit we're trying to achieve here except making code more verbose? We're using pretty much objects coming from the API. If the API changes, tests should fail anyway. We can't rely on devs reading console outputs, which is the only thing props validation really adds. I really can't see why we should add more verbosity to the calling code and the tests just because we can.

I'm open to consider doing this where it makes sense, for example in `AuthorBar` the properties used from `content` object kind of make sense as individual attributes. But in some of the receiving components they stop making sense individually once extracted from the `content`. For example, if we want to pass in an attribute called `id`, then the component name should say what the ID is. I don't particularly like the idea of generating extra verbose code and having longer component names just to get dev console warnings.

I'm not saying we should not add props validation when we have clear attributes that we pass in, but if it makes *sense* to pass in an object, we shouldn't add extra verbosity by ripping the object apart creating confusing out of context attributes on the receiving component.

jaywink: To get forward with this, I'm going to merge once tests pass. We can think of using the `v-bind` case by case according to above notes. In these cases I don't think it makes sense because:

* it increases verbosity with little benefit (only a console warning, having tests fail would actually be a little more useful)
* it can provide confusing variable names to local data
* it doesn't even allow validation objects with sub-objects, for example when passing `v-bind="content"` our `props` would still have to have `author: object`

Still, thanks for raising this! Let's keep the option case by case 👍

PR #362, opened 2017-11-11
Add load more trigger to StreamElement

jaywink:
When processing content in stream store, add a load more trigger to the fifth last content object. For the content that has this flag, render a load more div in StreamElement, which reacts to scrolling.

Fires an event to turn off the load more once content is scrolled to.

Refs: #202

TODO:
* [x] Add tests

@christophehenry, this solution seems to work, like I suggested in #346 comments. It will only fire the load more trigger once. Currently it doesn't have the dispatch to actually load more, just the trigger part. Sorry it wasn't really possible to send a PR to your branch any more.

I propose I finalize this (tests) and then merge it, then the actual data fetch can be added, the remaining parts in your PR? Might be easier to pick the still relevant parts into a new branch instead of trying to rebase the code. Sorry I couldn't help with this earlier.


christophehenry:
> I propose I finalize this (tests) and then merge it, then the actual data fetch can be added, the remaining parts in your PR?

No, please let me handle it. I want to avoid a painful merge.

christophehenry:
Which version of Firefox are you using? This solution doesn't work on Firefox 58 dev edition.
**Edit**: NVM. Found the solution.

jaywink:
Right, I think we have enough tests now. Managed to even find a bug while writing them :P Profile streams would have been totally broken.

Will merge once Travis is happy so we can move on to other things!

jaywink:
Thanks @christophehenry for your contribution here <3

PR #373, opened 2017-11-19
Fix profile name not appearing in Vue profile streams

christophehenry:
Failing Django tests. @jaywink Need halp!

halp_2

### DIFF
--- a/socialhome/streams/app/components/stamped_elements/ProfileStampedElement.vue
+++ b/socialhome/streams/app/components/stamped_elements/ProfileStampedElement.vue
@@ -59,7 +59,7 @@
                     <i class="fa fa-user" />
                     <i class="fa fa-arrow-right" />
                     <i class="fa fa-users" />
-                    &nbsp;{{ profile.followingCount }}
+                    &nbsp;{{ profile.following_count }}
                 </b-button>
             </div>
             <div class="mt-1">
@@ -72,15 +72,15 @@
                     <i class="fa fa-users" />
                     <i class="fa fa-arrow-right" />
                     <i class="fa fa-user" />
-                    &nbsp;{{ profile.followersCount }}
+                    &nbsp;{{ profile.followers_count }}
                 </b-button>
             </div>
         </div>
         <div class="d-inline-block">
             <img
-                v-if="profile.saferImageUrlLarge"
+                v-if="profile.image_url_large"
                 class="profile-stream-stamped-image"
-                :src="profile.saferImageUrlLarge"
+                :src="profile.image_url_large"
             />
         </div>
         <div class="d-inline-block ml-3 align-center stamped-profile-info">
@@ -88,7 +88,7 @@
             <h3><cite :title="translations.userHandle">{{ profile.handle }}</cite></h3>
         </div>
         <div class="text-center">
-            <b-button v-if="profile.pinnedContentExists" :variant="pinnedContentVariant" :href="urls.pinnedContent">
+            <b-button v-if="profile.has_pinned_content" :variant="pinnedContentVariant" :href="urls.pinnedContent">
                 {{ translations.pinnedContent }}
             </b-button>
             <b-button :variant="profileContentVariant" :href="urls.profileAllContent">
@@ -110,10 +110,10 @@ export default Vue.component("profile-stamped-element", {
             return this.profile.name ? this.profile.name : this.profile.guid
         },
         pinnedContentVariant() {
-            return this.profile.streamType === "pinned" ? "primary" : "secondary"
+            return this.profile.stream_type === "pinned" ? "primary" : "secondary"
         },
         profileContentVariant() {
-            return this.profile.streamType === "all_content" ? "primary" : "secondary"
+            return this.profile.stream_type === "all_content" ? "primary" : "secondary"
         },
         profile() {
             return this.$store.state.applicationStore.profile
--- a/socialhome/users/serializers.py
+++ b/socialhome/users/serializers.py
@@ -1,6 +1,8 @@
 from enumfields.drf import EnumField
+from rest_framework.fields import SerializerMethodField
 from rest_framework.serializers import ModelSerializer
 
+from socialhome.content.models import Content
 from socialhome.enums import Visibility
 from socialhome.users.models import User, Profile
 
@@ -35,13 +37,19 @@ class LimitedProfileSerializer(ModelSerializer):
 
 
 class ProfileSerializer(ModelSerializer):
+    followers_count = SerializerMethodField()
+    following_count = SerializerMethodField()
+    has_pinned_content = SerializerMethodField()
     visibility = EnumField(Visibility, lenient=True, ints_as_names=True)
 
     class Meta:
         model = Profile
         fields = (
+            "followers_count",
+            "following_count",
             "guid",
             "handle",
+            "has_pinned_content",
             "home_url",
             "id",
             "image_url_large",
@@ -55,8 +63,11 @@ class ProfileSerializer(ModelSerializer):
             "visibility",
         )
         read_only_fields = (
+            "followers_count",
+            "following_count",
             "guid",
             "handle",
+            "has_pinned_content",
             "home_url",
             "id",
             "image_url_large",
@@ -65,6 +76,16 @@ class ProfileSerializer(ModelSerializer):
             "is_local",
             "url",
         )
+
+    def get_following_count(self, obj):
+        return obj.following.count()
+
+    def get_followers_count(self, obj):
+        return Profile.objects.followers(obj).count()
+
+    def get_has_pinned_content(self, obj):
+        user = self.context.get("request").user
+        return Content.objects.profile_pinned(obj, user).exists()
 
 
 class UserSerializer(ModelSerializer):
--- a/socialhome/users/tests/test_views.py
+++ b/socialhome/users/tests/test_views.py
@@ -101,7 +101,6 @@ class TestProfileDetailView(SocialhomeTestCase):
 
     def test_get_json_context(self):
         request, view, contents, profile = self._get_request_view_and_content(create_content=False)
-        view.followers_count = 3
         self.assertEqual(
             view.get_json_context(),
             {
@@ -111,17 +110,25 @@ class TestProfileDetailView(SocialhomeTestCase):
                 "profile": {
                     "id": profile.id,
                     "guid": profile.guid,
-                    "followersCount": 3,
-                    "followingCount": str(profile.following.count()),
+                    "followers_count": 0,
+                    "following_count": profile.following.count(),
                     "handle": profile.handle,
-                    "saferImageUrlLarge": profile.safer_image_url_large,
-                    "streamType": view.profile_stream_type,
-                    "pinnedContentExists": view.pinned_content_exists,
+                    "has_pinned_content": Content.objects.profile_pinned(profile, request.user).exists(),
+                    "home_url": profile.home_url,
+                    "image_url_large": profile.image_url_large,
+                    "image_url_medium": profile.image_url_medium,
+                    "image_url_small": profile.image_url_small,
+                    "is_local": profile.is_local,
+                    "location": profile.location,
+                    "name": profile.name,
+                    "nsfw": profile.nsfw,
+                    "stream_type": view.profile_stream_type,
+                    "url": profile.url,
+                    "visibility": str(profile.visibility).lower(),
                 },
             },
         )
         request, view, contents, profile = self._get_request_view_and_content(anonymous_user=True, create_content=False)
-        view.followers_count = 3
         self.assertEqual(
             view.get_json_context(),
             {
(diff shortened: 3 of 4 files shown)


socialhome/users/views.py
```diff
@@ -67,18 +68,11 @@ def get_context_data(self, **kwargs):
...
-                "followersCount": self.followers_count,
-                "followingCount": str(self.target_profile.following.count()),
-                "handle": self.target_profile.handle,
-                "saferImageUrlLarge": self.target_profile.safer_image_url_large,
-                "streamType": self.profile_stream_type,
-                "pinnedContentExists": self.pinned_content_exists,
-            },
-        })
+        if self.request.user.is_authenticated:
+            json_context.update({"profile": ProfileSerializer(self.target_profile).data})
```
jaywink: I guess you're right about using the profile serializer to serialize the data for the frontend, but the trouble with adding all the extra fields there is that it will make the profile API a lot slower. But, since we disabled profile list API view totally some time ago, I guess performance isn't really an issue.

However, using `LimitedProfileSerializer` for unauthenticated users is wrong. It's not meant to control visibility, but to contain less data, ie the data required to render a grid item. A serializer doesn't itself do anything about visibility - that is what the view is for. If using a serializer directly, one must ensure only those fields are coming out that the user can see.

Our `ProfileSerializer` contains only data that is available always if the user can see the profile, so this is ok. Just remove the `else` and the usage of the limited serializer.

socialhome/users/serializers.py
```diff
@@ -51,10 +60,14 @@ class Meta:
             "location",
             "name",
             "nsfw",
+            "pinned_content_exists",
+            "safer_image_url_large",
```
jaywink: No need for `safer_*` image url method usage any more - those will soon be removed, since #348. And the API already has `image_url_large`.

socialhome/users/serializers.py
```diff
@@ -63,9 +76,20 @@ class Meta:
             "image_url_medium",
             "image_url_small",
             "is_local",
+            "pinned_content_exists",
+            "safer_image_url_large",
             "url",
         )
 
+    def get_following_count(self, obj):
+        return str(obj.following.count())
```
jaywink: No need for `str` - a count should be a number.

socialhome/users/serializers.py
```diff
@@ -63,9 +76,20 @@ class Meta:
...
+
+    def get_followers_count(self, obj):
+        return str(Profile.objects.followers(obj).count())
```
jaywink: No need for `str` - a count should be a number.

socialhome/users/serializers.py
```diff
@@ -63,9 +76,20 @@ class Meta:
...
+
+    def get_pinned_content_exists(self, obj):
+        return Content.objects.profile_pinned(obj, obj.user).filter(pinned=True).exists()
```
jaywink: `obj.user` should be the user *viewing the page*, not the user linked to profile. The `Content` queryset uses this to check what content is visible. For example, the profile might have pinned content which is not visible to the viewing user.

See [this serializer method in Content serializer](https://github.com/jaywink/socialhome/blob/bf8dc7af5a4cfe4f4d8e73aa6aa2464446818961/socialhome/content/serializers.py#L83) on how to get the user from the request object.

jaywink: No need for `pinned=True`, the `profile_pinned()` queryset method already does that.

socialhome/users/serializers.py
```diff
@@ -51,10 +60,14 @@ class Meta:
             "location",
             "name",
             "nsfw",
+            "pinned_content_exists",
```
jaywink: I would prefer maybe `has_pinned_content` but either way is fine.

socialhome/users/serializers.py
```diff
@@ -20,6 +22,7 @@ class Meta:
             "image_url_small",
             "is_local",
             "name",
+            "safer_image_url_large",
```
jaywink: No need for this as commented that we shouldn't use this serializer at all.

jaywink (changes requested):
Some things to tweak. Thanks for working on these!

Commit message should be a bit more clear on what is done, ie the refactoring of view json -> serializer would be nice to be mentioned. Also you're adding several properties to the Profile API which should be mentioned. Also, commits should reference an issue like this:

`Refs: #202`. For good commit messages, see https://chris.beams.io/posts/git-commit/

jaywink:
Oh I think the test wont crash any more if you pass in the correct user to `profile_pinned()` 👍

christophehenry:
> For good commit messages, see https://chris.beams.io/posts/git-commit/

I'm not good at writing informative git commit messages. 😢

jaywink:
> I'm not good at writing informative git commit messages. 😢

.. but learning new things or to be better at something is why we have these reviews!

socialhome/users/views.py
christophehenry: > But, since we disabled profile list API view totally some time ago, I guess performance isn't really an issue.

I prefer using one unique source rather than multiple. If there's a serializer for the API, better using it. What the front-end needs, any other application may need it. So, is it a go or a no-go?

jaywink: Sure, but we must also think of performance. As said, this is fine, since we only fetch individual objects, not lists. If we do in the future, we'll need to precalculate the values on the model. For now, this is fine.

socialhome/users/tests/test_views.py
```diff
@@ -387,7 +394,7 @@ def test_get_json_context(self):
         )
         request, view, contents, profile = self._get_request_view_and_content(anonymous_user=True, create_content=False)
         view.followers_count = 3
-        self.assertEqual(
+        self.assertDictEqual(
```
jaywink: This change is unnecessary - `self.assertEqual` calls `self.assertDictEqual` when comparing dict.

jaywink (changes requested):

socialhome/users/serializers.py
```diff
@@ -51,10 +60,13 @@ class Meta:
             "location",
             "name",
             "nsfw",
+            "has_pinned_content",
```
jaywink: Please place alphabetically.

socialhome/users/serializers.py
```diff
@@ -63,9 +75,20 @@ class Meta:
             "image_url_medium",
             "image_url_small",
             "is_local",
+            "has_pinned_content",
```
jaywink: Please place alphabetically.

socialhome/users/serializers.py
```diff
@@ -63,9 +75,20 @@ class Meta:
...
 
+    def get_following_count(self, obj):
+        return obj.following.count()
+
+    def get_followers_count(self, obj):
+        return Profile.objects.followers(obj).count()
+
+    def get_has_pinned_content(self, obj):
+        user = self.context.get("request").user
+        return Content.objects.profile_pinned(obj, user).filter().exists()
```
jaywink: `.filter()` is unnecessary here.

socialhome/users/tests/test_views.py
```diff
@@ -131,12 +138,12 @@ def test_get_json_context(self):
...
-                    "followingCount": str(profile.following.count()),
+                    "followers_count": "0",
+                    "following_count": str(profile.following.count()),
                     "handle": profile.handle,
-                    "saferImageUrlLarge": profile.safer_image_url_large,
-                    "streamType": view.profile_stream_type,
-                    "pinnedContentExists": view.pinned_content_exists,
+                    "safer_image_url_large": profile.safer_image_url_large,
+                    "stream_type": view.profile_stream_type,
+                    "pinned_content_exists": view.pinned_content_exists,
```
jaywink: I don't think `view.pinned_content_exists` should exist any more since we use the serializer now?

socialhome/users/views.py
```diff
@@ -67,18 +68,11 @@ def get_context_data(self, **kwargs):
...
-                "guid": self.target_profile.guid,
-                "followersCount": self.followers_count,
-                "followingCount": str(self.target_profile.following.count()),
-                "handle": self.target_profile.handle,
-                "saferImageUrlLarge": self.target_profile.safer_image_url_large,
-                "streamType": self.profile_stream_type,
-                "pinnedContentExists": self.pinned_content_exists,
-            },
-        })
+        if self.request.user.is_authenticated:
```
jaywink: This should be removed AFAICT - we always want to serialize the profile.

socialhome/users/tests/test_views.py
christophehenry: It's used on the Vue side.

jaywink: >  <b-button v-if="profile.has_pinned_content" 

I think you renamed it already in this PR :) Also, I was referring to the Django View. So [this](https://github.com/jaywink/socialhome/blob/6559f3ece9c7658ea797d2e01726e5c0ec24770d/socialhome/users/views.py#L106) and [this](https://github.com/jaywink/socialhome/blob/6559f3ece9c7658ea797d2e01726e5c0ec24770d/socialhome/users/views.py#L36) can be removed.

socialhome/users/views.py
christophehenry: Yes, that's right.

socialhome/users/tests/test_views.py
christophehenry: Won't it break the jQuery code?

christophehenry:
Fixing remarks and tests done. Waiting for another review.

socialhome/users/tests/test_views.py
```diff
@@ -111,32 +110,49 @@ def test_get_json_context(self):
...
                 },
             },
         )
         request, view, contents, profile = self._get_request_view_and_content(anonymous_user=True, create_content=False)
-        view.followers_count = 3
         self.assertEqual(
             view.get_json_context(),
             {
-                "currentBrowsingProfileId": None,
+                "currentBrowsingProfileId": request.user.id,
```
jaywink: Test is failing. Maybe because you're using a user ID, not profile ID?

socialhome/users/tests/test_views.py
```diff
@@ -370,18 +386,27 @@ def test_get_json_context(self):
         self.assertEqual(
             view.get_json_context(),
             {
-                "currentBrowsingProfileId": profile.id,
+                "currentBrowsingProfileId": request.user.id,
```
jaywink: Same here?

jaywink: You can also probably remove the `view.followers_count = 3` from this test slightly above line 385.

socialhome/users/tests/test_views.py
```diff
@@ -390,18 +415,27 @@ def test_get_json_context(self):
         self.assertEqual(
             view.get_json_context(),
             {
-                "currentBrowsingProfileId": None,
+                "currentBrowsingProfileId": request.user.id,
```
jaywink: Same here

jaywink: You can also probably remove the view.followers_count = 3 from this test slightly above line 389.

socialhome/users/views.py
```diff
@@ -10,6 +10,7 @@
 from socialhome.streams.views import StreamMixin
 from socialhome.users.forms import ProfileForm, UserPictureForm
 from socialhome.users.models import User, Profile
+from socialhome.users.serializers import ProfileSerializer, LimitedProfileSerializer
```
jaywink: You can remove `LimitedProfileSerializer` import.

jaywink (changes requested):
A few more tweaks + fixing of failing tests

socialhome/users/tests/test_views.py
christophehenry: Shouldn't be the same?

jaywink: No, local users have a `User`. Remote users don't. But everybody has a `Profile`. So what you should be setting (assuming a user is logged in) is `request.user.profile_id`. If it's an anonymous user, then pass in `None`, since anonymous users don't have a profile.

christophehenry:
Done. Come again, I'm waiting you ! >:D

socialhome/streams/app/components/stamped_elements/ProfileStampedElement.vue
```diff
@@ -72,23 +72,23 @@
...
                     <i class="fa fa-user" />
-                    &nbsp;{{ profile.followersCount }}
+                    &nbsp;{{ profile.followers_count }}
                 </b-button>
             </div>
         </div>
         <div class="d-inline-block">
             <img
-                v-if="profile.saferImageUrlLarge"
+                v-if="profile.image_url_medium"
```
jaywink: Why switch to medium? Large is 300px which is what is currently in the profile streams. Medium is 100px which is too small imho.

socialhome/users/tests/test_views.py
jaywink: This doesn't seem right. If the test passes it's probably because the ID's happen to match, not because they will always match. `currentBrowsingProfileId` should reference a `Profile` object ID. Does the test pass if you set it as `request.user.profile_id`?

socialhome/users/tests/test_views.py
jaywink: Same here, should not be `user.id`. Also, what is weird that we're using `AnonymousUser`, so this test should probably fail :/ I'll have a quick look at why it's not.

jaywink (changes requested):
A few minor comments.

socialhome/users/tests/test_views.py
jaywink: Ah, this is why:

```
In [4]: AnonymousUser().id is None
Out[4]: True
```

Still, this would be better more explicit to say `"currentBrowsingProfileId": None,` as it was before.

socialhome/users/tests/test_views.py
jaywink: Ah, this is why:

```
In [4]: AnonymousUser().id is None
Out[4]: True
```

Still, this would be better more explicit to say `"currentBrowsingProfileId": None,` as it was before.

socialhome/users/tests/test_views.py
christophehenry: `AnonymousUser` has no attribute `profile_id`. If I put `None` here, the test passes. I suppose `request.user.id` is `None` when using `AnonymouseUser`.

jaywink: Yeah the test should just check for the values it expects so just compare to `None` 👍

jaywink (approved):

jaywink:
Thanks! <3

PR #376, opened 2017-11-21
Switch pyembed to a custom own fork

jaywink:
Until changes can be upstreamed. Adds a timeout to outbound calls and
possibility to set extra flags. We set the DNT flag for Twitter oembed's,
even though it likely has no value since we do backend calls. Still nice
to communicate as it's possible to do.

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -33,7 +33,6 @@ lxml
 opbeat
 Pillow
 psycopg2
-pyembed
 python-opengraph-jaywink
 pytz
 redis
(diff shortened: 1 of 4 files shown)


christophehenry:
> custom own fork

images
depp-crybaby1_3161949b
index
study-men-not-women-are-more-likely-to-cry-at-work-1107049-twobyone
the-cry-face
cry
adibpro

jaywink:
Hey, forks are the way open source works! I totally intend to push upstream the two (so far) additions I've made. Sometimes, like with "markdownx", the changes are not accepted. Then the forks just live separately. It's open source. Don't cry, mon amie.

PR #378, opened 2017-11-25
Allow using email to log-in

### LINKED ISSUE #377: Login - allow email address in addition to nickname as login id
During the registration process we create an account with email and choose a nickname. 

I'm using Kee extension to register my password and it has detected my e-mail address has login. So on the first login I had an error about wrong password.

In reality it was because i try to use my email address to login instead of my nickname.

Having both nickname and e-mail address should avoid this kind of issue.

jaywink:
Closes #377


christophehenry:
Wow... That was a huge complex change. You should have talked about this before changing that many lines! :D

PR #381, opened 2017-11-26
Fetches new content when 'new posts' label is clicked by user

christophehenry:
Supersedes #360

- Changed logic of 'streamStoreOperations.newContentAck' to fetch new content
- Changed logic of 'streamStoreOperations.receivedNewContent' to stack received post ids to fetch in a list
- Replaced 'streamStore.state.hasNewContent' by 'streamStore.state.unfetchedContentIds' to stack content ids to fetch
- Render new posts when they are fetched
- Jump to top of page after content was fetched
- Tests ok

### DIFF
--- a/socialhome/streams/app/components/Stream.vue
+++ b/socialhome/streams/app/components/Stream.vue
@@ -30,15 +30,20 @@
 <script>
 import Vue from "vue"
 import imagesLoaded from "vue-images-loaded"
+import VueScrollTo from "vue-scrollto"
+
 import "streams/app/components/StreamElement.vue"
 import PublicStampedElement from "streams/app/components/stamped_elements/PublicStampedElement.vue"
 import FollowedStampedElement from "streams/app/components/stamped_elements/FollowedStampedElement.vue"
 import TagStampedElement from "streams/app/components/stamped_elements/TagStampedElement.vue"
 import ProfileStampedElement from "streams/app/components/stamped_elements/ProfileStampedElement.vue"
 import "streams/app/components/LoadingElement.vue"
+
 import {newStreamStore, streamStoreOperations} from "streams/app/stores/streamStore"
 import applicationStore from "streams/app/stores/applicationStore"
 
+
+Vue.use(VueScrollTo)
 
 export default Vue.component("stream", {
     store: newStreamStore({modules: {applicationStore}}),
@@ -86,7 +91,9 @@ export default Vue.component("stream", {
             Vue.redrawVueMasonry()
         },
         onNewContentClick() {
-            this.$store.dispatch(streamStoreOperations.newContentAck)
+            this.$store.dispatch(streamStoreOperations.newContentAck).then(
+                () => this.$nextTick( // Wait for new content to be rendered
+                    () => this.$scrollTo("body")))
         },
     },
     beforeCreate() {
--- a/socialhome/streams/app/components/Stream.vue
+++ b/socialhome/streams/app/components/Stream.vue
@@ -1,10 +1,10 @@
 <template>
     <div>
         <div class="container-flex">
-            <div v-show="$store.state.hasNewContent" class="new-content-container">
+            <div v-show="$store.getters.hasNewContent" class="new-content-container">
                 <b-link @click.prevent.stop="onNewContentClick" class="new-content-load-link">
                     <b-badge pill variant="primary">
-                        {{ $store.state.newContentLengh }} new posts available
+                        {{ translations.newPostsAvailables }}
                     </b-badge>
                 </b-link>
             </div>
@@ -70,6 +70,15 @@ export default Vue.component("stream", {
             } else {
                 console.error(`Unsupported stream name ${this.$store.state.streamName}`)
             }
+        },
+        translations() {
+            const ln = this.unfetchedContentIds.length
+            return {
+                newPostsAvailables: ngettext(`${ln} new post available`, `${ln} new posts available`, ln),
+            }
+        },
+        unfetchedContentIds() {
+            return this.$store.state.unfetchedContentIds
         },
     },
     methods: {
--- a/socialhome/streams/app/stores/streamStore.js
+++ b/socialhome/streams/app/stores/streamStore.js
@@ -5,6 +5,7 @@ import Vapi from "vuex-rest-api"
 import ReconnectingWebSocket from "reconnecting-websocket"
 import _defaults from "lodash/defaults"
 import _get from "lodash/get"
+import _pullAll from "lodash/pullAll"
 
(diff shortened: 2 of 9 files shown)


docs/changelog.rst
```diff
@@ -69,6 +69,8 @@ Added
 
   The logo also comes in a light version, for dark backgrounds. See :ref:`brand` for details.
 
+* Vue streams: streams load more content on scroll (`#346 <https://github.com/jaywink/socialhome/pull/346>`_)
```
jaywink: No need for changelog entries for Vue stuff since it's not a feature yet - we'll just have one changelog entry when released 👍

socialhome/streams/app/components/Stream.vue
```diff
@@ -71,6 +71,16 @@ export default Vue.component("stream", {
                 console.error(`Unsupported stream name ${this.$store.state.streamName}`)
             }
         },
+        translations() {
+            return {
+                newPostsAvailables: interpolate(
```
jaywink: `newPostsAvailables` -> `newContentAvailable`.

jaywink: We should avoid using globals (interpolate) when we can. How about `_.template` which can be imported? https://lodash.com/docs/4.17.4#template

```
>>> var t = _.template(ngettext("<%= ids %> new post available", "<%= ids %> new posts available", 3))
>>> t({ids: 5})
"5 new posts available"
```

jaywink (commented):
A few early notes 👍

socialhome/streams/app/components/Stream.vue
christophehenry: Why not?

jaywink: You mean why avoid? If there are options, the components should not IMHO rely on global javascript being available. This `interpolate` comes from Django right? Maybe at some point we want to get rid of the Django dependency (`gettext`). At that point, it's more work to convert these places. So as far as we're able to, let's make components that don't depend on magically available code.

But, thinking about this for a while, in this particular case maybe `interpolate` is fine. As long as we only use it when it's not so easy to use standard JS stuff or proper templates like `_.template`.

christophehenry: > Why not?

No, I meant: "*Yeah, why not use `_.template`*?". I was basically agreeing ;)

socialhome/streams/app/stores/streamStore.js
```diff
@@ -57,6 +58,14 @@ function fetchSharesSuccess(state, payload) {
     })
 }
 
+function fetchNewContentSuccess(state, payload) {
+    // It's important content is added before it's id is appended to `contentIds`
+    // to prevent it to be rendered when it's not defined yet
+    Vue.set(state.contents, payload.data.id, payload.data)
+    state.contentIds.unshift(payload.data.id)
```
christophehenry: Hmm... I didn't catch it but we may have a problem with this method. Content is added in the order it it fetched, not in the order id were received. Let's say `unfetchedContentIds` is `[34, 36, 39, 73]`. If id 73 is fetched before the other, we may end up with content being added in order `[73, 34, 36, 39]` to the top of the stream. Is it a bug or is it ok ?

christophehenry: Hmm... No, it is a bug. In the case of an empty stream, it modifies the logic of fetching more content as `last_id` should be 73 and not 39. Which means 73 would be fetched again.

jaywink: Well, `last_id` should always be the smallest `through` in the stream, because when we scroll we're scrolling backwards in time. So I don't think that applies here as the push notifications for new content will always be newer than anything currently in the stream (= higher `through` id's).

But still, how it *should* work, and how it works currently, is that newest content is at the top. Say you pull in 30 items to the top of the stream. They should be time ordered so that the newest ones are at the top.

I'm actually thinking we should probably think of not pulling new content using the Content API but the stream API. Something like the following:

* Instead of pushing "there is new content X" via websockets and storing a list of unfetched ID's, we would just store a "amount of new content on server", which would be updated by server.
* To pull new content (= user clicks the badge), we do an inverse fetch to the backend, so not `last_id` but `first_id` or something. This makes the server return everything (or X items) *above* the `first_id`.

How would this sound? I think it's doable on the backend, not sure how easily though. It would be better performance for frontend fetching and would easily solve the ordering issue, since the backend would return content in the right order, just like the load more functionality does.

Of course we don't have this backend API, so maybe just go with the current solution for now and pull in content according to the ID's, ignoring the ordering (and writing too many tests). How does that sound?

christophehenry: Hmm... Don't know. I always think about bad network coverage when it comes to fetching content. As much as possible, I prefer fetching atomic content one by one instead of one big request. It removes the single point of failure. If one request fail, at least you have the other content. But I admit it comes with nasty algorithmic problems. We could get back to the getter. Though I have no idea what performance cost it has. I don't know what to do :(

jaywink: It's cheaper to fetch a list of content in one go, just like you do when initially loading the stream. There is a lot of overhead doing separate requests. There is also a lot of overhead processing the payload in Vue, especially with older phones.

If the request fails, the client should retry a certain amount of times or at least tell the user to try in a bit.

Of course we shouldn't fetch hundreds of items. I propose to fetch X amount so one can kind of "scroll the stream" upwards. That would actually be quite unique? :) I think some Twitter clients I've used have that option.

socialhome/streams/app/components/Stream.vue
christophehenry: I used JS' template strings in the end.

socialhome/streams/app/stores/streamStore.js
christophehenry: New code fixes this behaviour.

socialhome/streams/app/stores/streamStore.js
```diff
@@ -62,6 +63,13 @@ function fetchSharesSuccess(state, payload) {
     })
 }
 
+function fetchNewContentSuccess(state, payload) {
+    // It's important content is added before it's id is appended to `contentIds`
+    // to prevent it to be rendered when it's not defined yet
```
christophehenry: This comment can be removed.

christophehenry:
Done here. Ready for review.

christophehenry:
Added scrolling top after content is fetched.

jaywink (approved):
Awesome, thanks! 👍 Let's bring this in for live testing 💥

christophehenry:
Yaye! @jaywink is back in the game!

PR #383, opened 2017-11-26
Vue streams quick reply in stream

jaywink:
* Add quick reply editor to grid replies container

  On clicking reply, don't open the full editor, instead show a quick
  reply box under the current replies. Save button adds the reply to the
  current replies.
  
  Provide a link to the full editor also since it has preview and image uploads.

* Bump bootstrap-vue to latest version.
* Make Content API parent writable

  Required to create replies via the API.

peek 2017-11-26 22-01

TODO:
- [x] Fix tests
- [x] Add tests for new functionality
- [x] Add missing validation logic to Content API parent/visibility as per todo's in serializer
- [x] Add changelog entry for Content API change
- [x] Remove hard coded visibility from vue side post once API doesn't require it for replies
- [x] Ensure everything works nicely for replies on shares
- [x] Add notes to API docs

Refs: #202

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -32,6 +32,10 @@ Changed
 
 * When processing a remote share of local content, deliver it also to all participants in the original shared content and also to all personal followers. (`#206 <https://github.com/jaywink/socialhome/issues/206>`_)
 
+* Allow creating replies via the Content API.
+
+  Replies are created by simply passing in a ``parent`` with the ID value of the target Content. It is not possible to change the ``parent`` value for an existing reply or root level Content object once created.
+
 0.6.0 (2017-11-13)
 ------------------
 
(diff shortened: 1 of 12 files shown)


christophehenry:
Add Markdown editor, pliiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiz :)

jaywink:
> Add Markdown editor, pliiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiz :)

I think this makes sense once the full editor is ported to Vue. It's a bit more work since it needs hooks to do the preview from the backend and do file uploads by button + drag'n'drop. We can then re-use the same components here in the stream 👍

jaywink:
API docs coming up, after that merge if no complaints!

Issue #338, opened 2017-10-20
Clicking on the comments icon in the stream should indicate loading


Vir-Cotto:
In the stream UI, when you click on the envelope to load comments (replies), there's no visual indicator to know that your action has been registered and the comments are loading, which can be a problem if there's a delay.

An icon change for that state change would be nice. 

How about the spinner? http://fontawesome.io/icon/spinner/ to indicate loading?

jaywink:
Definitely, thanks for logging! I'll add this to #202 since that is getting close enough to release that it might not make sense to add to the current soon legacy streams code.

The loading of replies is currently a bit buggy because it uses the websockets to fetch them. This has proven a bit unreliable which is why the new streams will use the Content API via ajax calls. Still, a spinner is definitely a good thing to have 👍

jaywink:
Added in Vue streams which will soon be default, PR #357 , closing.

PR #389, opened 2017-12-03
Fix pip-tools requirement per #338

ghost:
This adds pip-tools to requirements-dev.in (and then recompiled the list)


jaywink (approved):
Thanks! 👍 🎉

PR #390, opened 2017-12-05
Add Vue single content view

jaywink:
Make ContentDetailView render the Vue streams app if vue preference is set.
Tweak streams app to render a single content in case the stream name is
content stream. Some small refactoring to optimize grid stream related
actions only when rendering a grid. Small optimization to store in
preparing a 'stream' object when initializing the app. This will allow to
update the state according to routing events in the future.

Refs: #202

TODO:
* [x] Add test coverage for sigle content vue view

### DIFF
--- a/socialhome/streams/app/components/ReactionsBar.vue
+++ b/socialhome/streams/app/components/ReactionsBar.vue
@@ -34,7 +34,7 @@
             </b-button>
             <b-button v-else variant="secondary" @click.prevent.stop="share">{{ translations.share }}</b-button>
         </div>
-        <div v-if="showRepliesBox || $store.state.stream.single">
+        <div v-if="showRepliesContainer">
             <replies-container :content="content" />
         </div>
     </div>
@@ -59,6 +59,9 @@ export default Vue.component("reactions-bar", {
     computed: {
         showReplies() {
             return this.$store.state.applicationStore.isUserAuthenticated || this.content.reply_count > 0
+        },
+        showRepliesContainer() {
+            return this.showRepliesBox || this.$store.state.stream.single
         },
         showShares() {
             return this.$store.state.applicationStore.isUserAuthenticated || this.content.shares_count > 0
--- a/socialhome/streams/app/components/ReactionsBar.vue
+++ b/socialhome/streams/app/components/ReactionsBar.vue
@@ -34,7 +34,7 @@
             </b-button>
             <b-button v-else variant="secondary" @click.prevent.stop="share">{{ translations.share }}</b-button>
         </div>
-        <div v-if="showRepliesBox">
+        <div v-if="showRepliesBox || $store.state.stream.single">
             <replies-container :content="content" />
         </div>
     </div>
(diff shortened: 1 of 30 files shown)


socialhome/streams/app/components/ReactionsBar.vue
```diff
@@ -34,7 +34,7 @@
             </b-button>
             <b-button v-else variant="secondary" @click.prevent.stop="share">{{ translations.share }}</b-button>
         </div>
-        <div v-if="showRepliesBox">
+        <div v-if="showRepliesBox || $store.state.stream.single">
```
christophehenry: Hmm... Better put the condition in a plain function IMHO.

socialhome/streams/app/components/Stream.vue
```diff
@@ -1,35 +1,41 @@
...
+                <b-badge pill variant="primary">
+                    {{ translations.newPostsAvailables }}
+                </b-badge>
+            </b-link>
+        </div>
+        <div v-if="this.$store.state.stream.single">
+            <stream-element
+                class="grid-item grid-item-full"
+                :content="$store.state.content"
+                :key="$store.state.content.id"
```
christophehenry: `:key` is useless here has you do not iterate on a collection. (See [`key`](https://vuejs.org/v2/guide/list.html#key))

socialhome/streams/app/stores/streamStore.state.js
```diff
@@ -2,21 +2,37 @@ import _get from "lodash/get"
...
+    const content = _get(window, ["context", "content"], undefined)
     const contentIds = []
     const unfetchedContentIds = []
     const contents = {}
+    if (content) {
+        content.replyIds = []
+        content.shareIds = []
+        contentIds.push(content.id)
+        contents[content.id] = content
+    }
```
christophehenry, line 20: Why not use `contents` here and just define some `singleContentId` variable? This `singleContentId` can be replaced by a router later.

socialhome/streams/app/stores/streamStore.state.js
```diff
@@ -2,21 +2,37 @@ import _get from "lodash/get"
...
         contents,
         contentIds,
         hasNewContent: false,
         newContentLengh: 0,
         replies,
         shares,
-        showAuthorBar: streamName.length > 0 ? !streamName.startsWith("profile_") : false,
+        showAuthorBar: !stream.isProfile,
+        stream,
         streamName,
```
christophehenry: I think that can be removed, then.

christophehenry (changes requested):

socialhome/streams/app/stores/streamStore.state.js
jaywink: Yeah, good point, fixed.

socialhome/streams/app/stores/streamStore.state.js
jaywink: The stream name is still used in full also by at least the websocket url. Would be good to refactor things a bit, but leaving this in the state for now since refactoring that isn't really related to this pr.

jaywink:
Fixed remarks! If nothing to add, will merge once tests pass for more live testing.

jaywink:
Oh, also added a management command to add some basic dummy content, for #329 - not really related to but needed some content to test this since I reinstalled my machine :P

christophehenry:
Go on ! Merge that shit !

Issue #340, opened 2017-10-20
How to run pod on 512mb memory


autogestion:
Is it possible to turn off some of Socialhome features , without ruining its workability, and get as a result some light-weight server for limited range of users, which could be ran on cheap hosting

jaywink:
So basically the current approximate requirements are [here](https://socialhome.readthedocs.io/en/latest/installation.html#resources) - an estimate of 745mb with a default configuration. This is *excluding* the database and Redis. All in all we're talking somewhere over 1gb. We're increasing the amount of pre-calculated in memory stuff so memory requirement is likely to go up a bit.

But, all hope is not lost. The current defaults are not a minimal configuration. I think by decreasing uWSGI and RQ processes to minimum one, memory could be around 375mb, though not including database and redis. If you can run those and still keep it within 512mb, there is hope.

If RQ workers are decreased, then it is vital to turn off the relay functionality. One worker will not be enough to process all the incoming public content from the network. Basically a setting for [this](https://github.com/jaywink/socialhome/blob/e169ba4857f9e188d89391c1f5618100a0535cc1/socialhome/federate/views.py#L116).

Another thing that could be turned off by configuration (not yet, but with an additional setting) would be the streams pre-calculation. That would not only save memory, but cause less RQ jobs. It would make streams load slower though, and not generate a good user experience with complex stream filters.

Let's leave this issue open for discussing and improving the minimal memory footprint use case.

autogestion:
Could we get some speed, meaning greatly reduce the server load by using asynchronous framework instead of django. I'm talking about sanic/uvloop. It will speed up loading post that are already fetched from the fed, but could it speed up fetching/broadcast? theoretically

emacsen:
@autogestion I'm sure @jaywink will weigh in on this but as a casual user of the software, I see two things going on here:

1. Would changing to another framework improve performance?

2. Would it be feasible to change web frameworks?

I think the answer to your first question (would it improve performance) is a question that's hard to answer without profiling. It's easy to say that one software is faster than another, but the real trick is to do the kind of detailed performance testing that would be necessary to see where the actual bottlenecks are. And then of course the question of performance is tied into the question of "Performance where?" Meaning is the experience of "speed" about the software, or network issues, the frontend UI code, etc? Most of the time, "speed" from the user's perspective had nothing to do with the experience of the server, which is the only thing that changing web frameworks could do.

[...]

PR #391, opened 2017-12-13
Docker config to run sh instance with ssl

autogestion:
certs added to commit are for localhost, taken somewhere from github, and were added in order to simple run-&-test this config


ghost:
This isn't my decision, but can you explain how this is better than a standard nginx docker config that proxies to django?

autogestion:
@Vir-Cotto Could you please provide an example of standard nginx docker config

ghost:
Taking a step back... There's really three situations I see:

A dev environment
A small/medium size production environment
A large production environment

For the dev environment, I wrote the included Docker example. It would be possible to use the Django built in SSL server but I omitted doing so because (as far as I know) it's not needed in texting.

For a large production environment, you'd be using a more complex setup, maybe a hardware load balancer and SSL accelerator- we need not think too hard about that since it's outside our scope.

In a medium size production environment, you're still going to have multiple servers and if you're using Docker, probably some kind of software provided network overlay, which your config doesn't cover.

So now we're down to the small server example.

And in that case, I'd be inclined to use something like this:
https://github.com/evertramos/docker-compose-letsencrypt-nginx-proxy-companion

(and by inclined, I mean I use this in production for some small things- not for Socialhome, but it could be adapted easily).

I think what you've provided is great, but the above example has some notable advantages, including:

1. It doesn't lock a service to a port (your example assumes that the target is at port 8000, which means that you'd need to do the same for any other service

2. There's an auto-configuration of services by name. That means you can drop new service in and it just works.

3. It doesn't embed the SSL cert in the Docker image (not sure why yours does but it seems to?)

4. It supports auto-reconfiguration

5. It supports an "add-on" in the form of LetsEncrypt:

https://github.com/evertramos/wordpress-docker-letsencrypt


So while I think this is a a great effort and I know that I'd love help with some production docker images (I think @jaywink  was working on them) , this is less of a good start than some of the things that are already out there.  That doesn't diminish your work, but maybe we can focus it somewhere where it would be more immediately helpful?

jaywink:
Thanks @autogestion (and @Vir-Cotto for comments) - will check during the weekend. Right now suffering from nasty case of flu which doesn't want to let go.

autogestion:
The reason I started to do this config was an effort to deploy locally 2 instances of sh (Long live the Docker), which could communicate with each other to test some ideas. sh instances could communicate only using ssl (that's hardcoded in federation library). So this config is the extension of @Vir-Cotto  dev config with ssl support, and it is not offered (but can be used, for shared hostings which not supports letsencrypt) as production config. They could be merged in one. I think

ghost:
If it's for development purposes, there is a python package whose name escapes me at the moment (perhaps https://github.com/teddziuba/django-sslserver ?)  that will allow for SSL from the built in server.

It's not performant, but it doesn't then require a second process or any other complex config.

ghost:
Thinking about this a bit more, some further thoughts:

Firstly, I think we'll want a solution that would for non-Docker users as Docker users. That's a vote for the Python SSL acceleation. Then we can modify the development Docker container to offer it- problem solved.

And then we'll probably want to include some tools to generate the cert. Maybe something like 
manage,py genselfsignedcert

autogestion:
I'm thinking about updating dev config with option to use django-sslserver (and option to use collectstatic instead of running npm instance). And updating this config so it could be used for small/medium size production environment (  with an eye on such cases https://github.com/jaywink/socialhome/issues/340 ) with tools to generate cert etc

ghost:
I think that updating the dev config makes a great deal of sense, and then having self-signed certificate tools there makes sense.

As for production, I think we should make that a separate discussion.

jaywink:
> sh instances could communicate only using ssl (that's hardcoded in federation library).

I think this should change. It was a mistake from my part to enforce HTTPS on the federation library level. Sure it should *default* to it if the given URL doesn't have protocol, but it should also allow passing in a HTTP url. I will change that so we don't need to use SSL for the use case you describe. So no need to hack on anything extra. We should just enhance the current Docker development images that exist in case they are missing something (except not SSL).

As said, production docker configuration is a completely different subject 👍

Thanks for the PR! I'll close it though as we should rather remove the federation constraint than duplicate docker configuration. Please expand the current dev configuration if something is missing there.

Logged #394 and jaywink/federation#106

PR #395, opened 2017-12-24
Fix profile dropdown problems in VueJS stream

christophehenry:
Fixes:

> - [ ] Bug: dropdown renders too much off to the left and is not wide enough for all the text:
> - [ ] Bug: Organize profile content link leads to email settings


jaywink (approved):
Great, thanks! 👍 🎉

Issue #322, opened 2017-10-14
[VueJS] Notifications


christophehenry:
With the Vue front-end going big, I think it's time to properly handle notifications. We have a few use-cases that need a proper error hadling. Hopefully, VueJS has tons of plugins including a bunch of notification libraries listed [here](https://github.com/vuejs/awesome-vue#notification). I've excluded libraries that received no commit for more than 3 months and those not proposing a proper demo to narrow the choice to:

* [Vue toasted](https://github.com/shakee93/vue-toasted)
* [Vue notification](https://github.com/euvl/vue-notification)
* [Vue snotify](https://github.com/artemsky/vue-snotify)

In my opinion, Vue snotify is the best. It's pretty, it's rather small and offers fancy features. But it's also opiniated regarding design.

What do you think?

jaywink:
So by notifications you mean things like "share action failed", not a notifications menu listing peoples actions like replies, shares and reactions?

If the latter then it's blocked until we have notifications in the backend too. Currently all notifications are just via email.

If you meant the former then yes, would be nice to have some information delivered to user on the event of an error happening, for example failing to load the stream or share content. I'll leave it up to you which library, but visually I would like it to fit into the current UI, so probably no huge boxes like in some of these demos. Colour is ok, but the notifications should be relatively simple and not "in your face" as the UI is simple too.

Thanks for looking into these! I'll add a link to this issue in #202 .

christophehenry:
#398 adding notifications is merged. Closing.

PR #398, opened 2017-12-28
Add notifications

christophehenry:
- [x] add notification system to Vue streams.
- [x] tests

@jaywink Styling can be changed later.

ref: #202 #322

### DIFF
--- a/socialhome/streams/app/components/ProfileReactionButtons.vue
+++ b/socialhome/streams/app/components/ProfileReactionButtons.vue
@@ -42,6 +42,9 @@
 </template>
 
 <script>
+import Vue from "vue"
+
+
 export default {
     name: "profile-reaction-buttons",
     data() {
@@ -58,6 +61,12 @@ export default {
         currentBrowsingProfileId() {
             return this.$store.state.applicationStore.currentBrowsingProfileId
         },
+        isUserAuthenticated() {
+            return this.$store.state.applicationStore.isUserAuthenticated
+        },
+        authorNameOrHandle() {
+            return this.profile.name ? this.profile.name : this.profile.handle
+        },
         showFollowBtn() {
             return this.$store.state.applicationStore.isUserAuthenticated && !this.following
         },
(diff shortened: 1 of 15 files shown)


webpack.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = {
                 test: /\.scss$/,
                 loaders: ["css-loader", "sass-loader"],
             },
+            {
+                test: /\.css$/,
```
jaywink: Hmm, won't this duplicate quite a bit?

christophehenry: This can be removed for now. Importing CSS files isn't working yet.

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -112,7 +122,13 @@ export default Vue.component("author-bar", {
             }
             this.$http.post(`/api/profiles/${this.currentBrowsingProfileId}/add_follower/`, {guid: this.author.guid})
                 .then(() => this.isUserFollowingAuthor = true)
-                .catch(err => console.error(err) /* TODO: Proper error handling */)
+                .catch(error => {
+                    const msg = "An error happened while trying to follow ${user}"
+                    const translated = gettext(msg)
+                    Vue.snotify.error(_template(translated)({user: this.authorNameOrHandle}))
+                    console.error(`${_template(msg)({user: this.authorNameOrHandle})}; error met: ${error}`)
```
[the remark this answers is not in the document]
jaywink: I think console logging is only useful for development or if we plan to do some better notification but don't have it yet. Now that this does, I think the console logger should be removed.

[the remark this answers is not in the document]
christophehenry: I'm not in favor of displaying the stacktrace to the user. But Logging it in console could help us in case of a nasty bug. Dunno. You decide.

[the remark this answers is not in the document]
jaywink: Definitely no stacktrace to users - it wont be helpful to anyone. But I'm not sure outputting a stacktrace to the console will help either. Users wont know to look there.

Having `console` calls can even crash the JS code on certain browsers, like [IE9 and below](https://caniuse.com/#feat=console-basic). Of course, we don't really care about such old browsers since our code likely will crash on them anyway, but still, good to keep in mind. Maybe this was the reason in Diaspora.

I'm in favour of generally not logging stuff to the console. But then we're in highly development mode. Maybe in some cases it makes sense to be able to see production errors until code is seen as more stable. I don't think this is such a place since the error is likely just going to be a backend error message or a network error. The former can be seen by the node admin. The latter can be probably seen from the network call panel, if needed. So, for this one, I would say remove it 👍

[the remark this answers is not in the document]
christophehenry: Duly noted. Removing `console` directives, then (and hopping nobody uses our beautiful application on that IE crap anyway).

christophehenry:
FINISHED! AT LAST!

socialhome/streams/app/components/ProfileReactionButtons.vue
```diff
@@ -72,25 +81,33 @@ export default {
...
     methods: {
         follow() {
-            this.$http.post(
-                Urls["api:profile-add-follower"]({pk: this.currentBrowsingProfileId}),
-                {guid: this.profile.guid})
-                .then(() => {
-                    this.following = true
-                })
-                .catch(err => console.error(err) /* TODO: Proper error handling */)
+            if (!this.isUserAuthenticated) {
```
jaywink, line 93: I think it's OK to add these, but it's unnecessary effort. We don't render the buttons if the user is not logged in. So, unless someone makes a mess of the code the user will never see this notification. And even if someone does, the backend will still validate and raise an error. Better have some kind of central error notify for these kinds of things rather than raising errors that will never be raised as the code cannot be reached.

Let's leave them in now as they are here, but for the future, to save your time ;)

jaywink (approved):
A few comments, but nothing stopping me merging this in - sorry for the wait, and thanks! <3 <3

socialhome/streams/app/components/ProfileReactionButtons.vue
christophehenry: > I think it's OK to add these, but it's unnecessary effort.

If shit can be done with your app, shit *will* be done with you app ;)

jaywink: Not on my watch... 😁

christophehenry: dd505ffbfd7b5dc7f7e0fd75fbc5e8cb17c8f722ce54ab1ba2386fbb3d31a6f8

PR #402, opened 2018-01-13
Vue profile stamped element improvements

jaywink:
A few things related to make the Vue profile stamped element work like the old one.

* Add follower/following counts for other than own profiles
* Fix alignment of profile actions dropdown
* Add profile buttons to non-own profile, including link to remote profile and follow/unfollow. Placed this code in a separate component which is now also reused in stream element author bar.

selection_004

selection_005

selection_006

TODO
* [x] Fix/improve tests
* [x] Add changelog for API addition
* [x] Make strings translatable

Ping @christophehenry for review.

Refs: #202


jaywink:
Ready!

socialhome/streams/app/components/AuthorBar.vue
```diff
@@ -50,8 +22,13 @@
 <script>
 import Vue from "vue"
 
+import ProfileReactionButtons from "./ProfileReactionButtons.vue"
+
 
 export default Vue.component("author-bar", {
+    components: {
+        ProfileReactionButtons,
```
christophehenry, line 30: Why is this needed? Is it because you don't use `Vue.component` on the component?

socialhome/streams/app/components/ProfileReactionButtons.vue
```diff
@@ -0,0 +1,104 @@
...
+            :title="translations.follow"
+            :aria-label="translations.follow"
+        >
+            <i class="fa fa-plus"></i>
+        </b-button>
+    </div>
+</template>
+
+<script>
+export default {
```
christophehenry, line 45: Why no `Vue.component`?

christophehenry:
Just made a quick and tired-mode review. I trust you to dump good code anyway 👍

socialhome/streams/app/components/AuthorBar.vue
jaywink: Yes. I noticed in December that the way we register components (`Vue.component`) makes it difficult to use `shallow` rendering in tests, which makes it hard to write tests for high level components since `mount` will render all the child components too. It might be a combination of things, but from a little experimentation after refactoring many of our component files I got some good results. Unfortunately it would have been a lot of work to clean up so I abandoned the refactor.

If ok, from now on, if we could favour exporting components as objects instead of calling `Vue.component`. This is a bit more verbose, since the components need to be registered in the components that use them (or globally some other way), but it also makes things a bit more visible on the component what child components it will require. Currently sometimes we don't even import the components we use in the template, since it's already registered, which is confusing.

From the Zen of Python:

> Explicit is better than implicit.

While global registering of components is supported, exporting just an object is in the Vue [single file components](https://vuejs.org/v2/guide/single-file-components.html) example. I couldn't find real evidence on which is suggested to be used, both seem to work quite fine, except Avoriaz seems to be a bit picky when doing shallow rendering.

christophehenry: Another option would be to export both the object and the Vue component. The Vue component as default and the object for the tests. I don't like having to manually register component. Looks error prone.

> Currently sometimes we don't even import the components we use in the template, since it's already registered, which is confusing.

Really? How can it work?

> except Avoriaz seems to be a bit picky when doing shallow rendering.

It's probable that this problem was fixed in the incoming [Vue test utils](https://vue-test-utils.vuejs.org/en/).

christophehenry: > global is bad! they said that in school!

At school, they also teach you to always write accessors, use the builder pattern and singleton are evil. So I suppose I can code without what they teach at school.

PR #403, opened 2018-01-14
Twitter streams OEmbed refactor

jaywink:
Fix (or hack around) how Masonry layout is done in Vue streams for Twitter OEmbed widgets.

* Load widget script once in base template
* Do a layout (hackily) after 2 and 4 seconds (if oembed's), to ensure Twitter widget has had time to load content
* Add `has_twitter_oembed` to Content API as a read only field

TODO:
* [x] ~~Check~~ Ensure layouting triggers after load more
  - Clear the flag on load more?
* [x] Trigger widgets script after Vue initialized... how and where from?
  - Maybe call `twttr.widgets.load()` after each mount of streamelement with twitter oembed content?

Refs: #202

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -38,6 +38,8 @@ Added
 
 * Installation docs now have an example SystemD service configuration, see :ref:`installation-other-systemd`. (`#397 <https://github.com/jaywink/socialhome/issues/397>`_)
 
+* Content API has a new read only field ``has_twitter_oembed``. This is ``true`` if the content text had a Tweet URL *and* a fetch for the OEmbed code has been successfully made.
+
 Changed
 .......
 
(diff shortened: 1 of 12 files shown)


christophehenry:
> Do a layout (hackily) after 2 and 4 seconds (if oembed's), to ensure Twitter widget has had time to load content

Indeed, it's not nice. Have you tried enclosing Twitter embed in an iFrame and use the [onresize event](https://www.w3schools.com/jsref/event_onresize.asp) to detect image loading?

jaywink:
> Have you tried enclosing Twitter embed in an iFrame and use the onresize event to detect image loading?

But we don't know the width or height of the content. I'm not a huge iframe fan or haven't used them a lot, how would that work?

Also it's not about just images - that is actually easier. The problem is normal tweets also since the twitter widget takes a little time to load the content and styles.

jaywink:
Should work better now. More hacks than before, but less hacky *looking* code! :)

christophehenry:
Still... Please log an issue to fix that later. If the network is slow the hack could be broken.

PR #415, opened 2018-01-25
Fix incorrect throughs values when collecting content in Streams

### LINKED ISSUE #412: resharing from diasp.org produced duplicates
After resharing some old posts from my diasp.org account,
my profile has duplicates of those reshares,
and also duplicates of content original to my socialhome account.

https://socialhome.network/u/dvs1/all/

jaywink:
Previously we set share through as content.id for any shares included in
streams via DB lookups. This caused a problem of missing content in new
streams and infinite load on old streams. By adding through to all content
querysets and then correctly populating it for shares, we can now produce
a correct through value for content included in streams through shares
in DB lookups too.

Fixes #412

### DIFF
--- a/socialhome/content/querysets.py
+++ b/socialhome/content/querysets.py
@@ -1,5 +1,5 @@
 from django.db import models
-from django.db.models import Q
+from django.db.models import Q, Min, F
 
 from socialhome.content.enums import ContentType
 from socialhome.enums import Visibility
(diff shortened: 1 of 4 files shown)


socialhome/content/querysets.py
```diff
@@ -81,7 +81,7 @@ def public(self):
         return self.top_level().filter(visibility=Visibility.PUBLIC)
 
     def shares(self, share_of_id, user):
-        qs = self.visible_for_user(user).filter(share_of_id=share_of_id)
+        qs = self.visible_for_user(user).filter(share_of_id=share_of_id).annotate(through=Min('id'))
```
christophehenry, line 84: 🎉

PR #427, opened 2018-02-11
LET'S NUKE THE OLD STREAM FRONT-END!

christophehenry:
Refs: #202


jaywink:
Thanks <3 Will try to check today 👍 Let's 🔥 some code!

jaywink:
Sorry still haven't been able to merge this. Looks good, just need to test a bit locally and more importantly log some tasks to the Vue epic about things we haven't implemented in the Vue side which need to be taken care of at some point - before they are now removed and never seen again ;)

jaywink:
OK let's merge 👍 💥 Some things might be unused now in the backend and I noticed we're not setting titles/og tags properly in vue pages - I logged follow up issues #428 and #429. But nothing in this PR looks dangerous enough not to move forward with the cleaning. Thanks!

PR #431, opened 2018-02-17
Remove Vue front-end parameter

christophehenry:
This one was forgotten by #427


christophehenry:
Wait, what?

jaywink:
> Wait, what?

Absolute coverage will go down if you remove code that was previously covered :P 

Good catch, thanks!

Issue #432, opened 2018-02-18
Rewrite front-end to Vue


christophehenry:
I'm logging this issue as a follow-up of #202. Now that Streams front-end is almost done, what is the next step?

jaywink:
I think the next step is something like the following:

1) Refactor Vue streams app into a generic frontend app (ie place it under socialhome app, not streams, refactor store, rename many things, etc)
2) Add routing in Vue
3) Add whatever components we need to replace the Django views

I'm not entirely sure how we can do this like we did the streams by using a flag and keeping old functionality working. Maybe if 1) is done first, the streams can still be loaded by creating two bundles, one for just streams functionality, one for new frontend. That would allow slowly building the new frontend and provide a flag to use it, without having to do all things at once quickly. The alternative is a painful to maintain feature branch which needs to have merges done constantly from master. I'm not a huge fan of this idea.

I would keep the frontend in the same repo still. Let's not add too much overhead to the small project :)

Did you have some other thoughts or are we aligned here?

christophehenry:
We're aligned.

christophehenry:
Now giving a second thought, I didn't foresee the whole Vue front would be in its own application. This means the the whole front has to be loaded at once since Webpack and ES6 don't manage conditional imports like Python does. I was expecting we could split the front-end JS code in multiple parts to reduce the amount of uselessly downloaded code. This will ultimately represent a shitload of JS code. We need to find a way to use browser caching at some point.

jaywink:
I don't see the size of the bundle a huge problem. Most of the code is already written. The streams are like 95% of Socialhome. AFAICT it has to be pretty much one bundle so we can make it route nicely in the JS side which makes page transitions faster since there is no page reloading. If we have many bundles and thus independent pages, the transitions are slower, just like currently.

The streams bundle is currently ~500kb or so. That is probably 1/20 of the amount of data used by images in a normal stream refresh. There is nothing we can do to make data usage small unless we start restricting local and remote image loading, which uses many many more times data than our JS does.

jaywink:
I mean if @christophehenry its something you want to do, you could rewrite every single page in vue as separate apps (and bundles), and then we merge them all later into one and add vue routing. Might even be easier than all at once, not sure. Just an option if you want to continue working on the vue side. Though, one could also add stuff to the streams, like filters and content actions.

christophehenry:
Ok. Let's do one bundle and see with time if frontend gets bloated.

[...]

PR #439, opened 2018-02-24
Move frontend in its own Django app

christophehenry:
Pretty much straightforward, but a requirement to adding Vue routing, which is the next step.
Ref: #432


jaywink (approved):
Good stuff, not much to add here! :tada: :beers: 

Thanks!

PR #441, opened 2018-02-24
Add Vue router

christophehenry:
Simple mockup of what can be done with VueRouter.
Needs tests, I guess. But first, I need to eat something.


christophehenry:
Ready for merge.

jaywink:
Will need to test this locally a bit. Not sure what actually changes with this code, is it just moving the stamped element logic to vue router? Does anything else change?

christophehenry:
> is it just moving the stamped element logic to vue router?

Yes, that's it. Things like tag and profile matching still uses infos in the store. It a bit trickier. I prefre a seperate PR.

socialhome/frontend/routes.js
```diff
@@ -0,0 +1,35 @@
+import VueRouter from "vue-router"
+import Stream from "frontend/components/streams/Stream.vue"
+
+
+function $$(props) {
+    return route => Object.assign({}, route.params, props)
+}
+
+const routes = [
```
jaywink: This is missing profile pinned (ie default profile view `/u/username/` and also profile `/p/guid/` and also single content `/content/id/shorttext/` - at least, tested locally they produce empty pages.

This is why we must have a basic test for each router is covering them. I guess higher level browser tests would be nice at some point. But for now some vue-router tests to replace the ones removed by this PR.

webpack.config.js
```diff
@@ -0,0 +1,2 @@
+// This file is for helping IntelliJ
+module.exports = require("./config/webpack/webpack.dev")
```
jaywink: This should be in the dev webpack config file.

jaywink (changes requested):

webpack.config.js
christophehenry: No. I stated in the comment, this file is only here to help IntelliJ with module import auto-completion. IntelliJ looks for this file at the root of the project and detects WP configuration.

socialhome/frontend/routes.js
christophehenry: Question: shall we use `Urls` too? Like for [Vapi](https://github.com/jaywink/socialhome/blob/master/socialhome/frontend/stores/streamStore.js#L82)?

christophehenry: I have a problem with `/`. I don't know how to match the dynamic stream.

socialhome/frontend/components/streams/Stream.vue
```diff
@@ -65,21 +66,6 @@ export default Vue.component("stream", {
...
-                    return "PublicStampedElement"
-                case "tag":
-                    return "TagStampedElement"
-                case "profile_all":
-                case "profile_pinned":
-                    return "ProfileStampedElement"
-                default:
-                    console.error(`Unsupported stream name ${this.$store.state.stream.name}`)
-            }
-        },
```
christophehenry: Sadly, I can't remove that part. It's necessary to match `/`.

socialhome/frontend/routes.js
jaywink: I don't think we should bother with `Url` here.. Too much overhead for the little benefit of not having to change urls in frontend too. They should not change that often.

About `/`, I have no idea... would need access to either `window` or the store to get what the root stream is..

webpack.config.js
jaywink: I think you can define it in settings? But sure, I don't mind. But could you clarify with the longer version of the comment like above in the file and also add a comment indicating where the real webpack config is?

socialhome/frontend/routes.js
christophehenry: Sadly, I don't think we can get rid of [this](https://github.com/jaywink/socialhome/pull/441/files#diff-f040736819fe5a79d0c27803dd925b36L68).

christophehenry:
Ok, this looks good for now. This is temporary, though. Some changes needs API to support GUID. Other need more deep refactoring. This PR is big enough IMHO. I have one line of code not covered which is [event emission here](https://github.com/jaywink/socialhome/pull/441/files#diff-7e25e2d4658f615fad5102b105889ce7R104). I don't think it's a big deal since I plan to quickly refactor [`loadStream`](https://github.com/jaywink/socialhome/pull/441/files#diff-f040736819fe5a79d0c27803dd925b36R106) in a [mixin](https://vuejs.org/v2/guide/mixins.html) that `StreamElement` can make use of.

Long story short: ready for review.

socialhome/frontend/components/streams/Stream.vue
```diff
@@ -31,25 +32,30 @@
...
+import FollowedStampedElement from "frontend/components/streams/stamped_elements/FollowedStampedElement.vue"
+import TagStampedElement from "frontend/components/streams/stamped_elements/TagStampedElement.vue"
+import ProfileStampedElement from "frontend/components/streams/stamped_elements/ProfileStampedElement.vue"
+import "frontend/components/streams/LoadingElement.vue"
+
 
 Vue.use(VueScrollTo)
 
 export default Vue.component("stream", {
+    // TODO: Seperate Stream.vue into TagStream.vue, GuidProfile.vue and UsernameProfile.vue, etc. in the future
```
jaywink, line 53: The `/u/username/` is not actually a real page, it just renders the real profile stream, if you look at what happens in the backend, so at least these should just use the same code. In general, I'm not sure we should start separating the code of different streams too much since other than the stamped element part they are identical in function.

jaywink (commented):
This has quite a bit of refactoring in it, will need to check in more detail and do also local testing. Will get back to this soon! :)

socialhome/frontend/components/streams/Stream.vue
christophehenry: Sure, but you have one difference which is the `props` attribute. `guid`, or `username` or `tagName`. Most of the code will be identical but we can use Vue's extension system to create a class hierarchy. Eventually, the stamped element's code must be merged into theses new stream classes along with the new Stream classes. It's a bit difficult to explain without throwing some code.

christophehenry:
So, what's the status on this ?

jaywink:
Still need to check this locally. Will try to do today!

socialhome/frontend/components/streams/Stream.vue
jaywink: As long as we avoid code duplication I'm fine :+1: 

There will be many new types of streams so I would want the code to stay unified to avoid having to change each stream file when adding say a filter or other minor thing.

socialhome/frontend/routes.js
```diff
@@ -0,0 +1,25 @@
...
+    // TODO: In the future, Stream can be replaced by a generic RootStream.vue that just delegates to another component
+    {path: "/", component: Stream},
+    {path: "/streams/followed", component: Stream},
+    {path: "/streams/public", component: Stream},
+    {path: "/u/:user", component: Stream, props: $$()},
+    {path: "/u/:user/all", component: Stream, props: $$()},
+    {path: "/p/:guid", component: Stream, props: $$()},
+    {path: "/p/:guid/all", component: Stream, props: $$()},
+    {path: "/streams/tag/:name", component: Stream, props: $$()},
+]
```
jaywink: Single content view (stream!) is missing, basically `/content/<id>/<shorttext>/`.

The short text is just eye candy (nicer urls), you can ignore it for routing, only the ID matters. Sometimes there is no short text at all for weird languages that can't be slugified, so you'll have to support also `/content/<id>/` style version.

Oh another comment, all the urls in Django have a postfix `/` which is missing from all the urls above. Maybe it doesn't matter here but would like to keep this style for consistency.

socialhome/frontend/stores/applicationStore.js
```diff
@@ -10,6 +10,7 @@ function newApplicationStore() {
         state: {
             isUserAuthenticated: _get(window, ["context", "isUserAuthenticated"], false),
             currentBrowsingProfileId: _get(window, ["context", "currentBrowsingProfileId"]),
+            // TODO: To be removed
```
jaywink, line 13: I've been thinking a lot about the stream loading time which is a big usability thing for me. We had quite a big regression on this with the Vue streams since content is not embedded into the page by Django. I would like to re-introduce that so that the initial page load would add the first set of content and then Vue would handle all the fetching after that (load more, switching streams, etc). This would probably cut down quite a bit of seeing the spinner. It's a nice spinner but still I would like to see it less :)

Thus I think the profile info should also stay, no real need to remove it. The only thing we would be adding is an additional network call on the initial request. Of course Vue should update the information as need be, but I'm just talking about the initial load which we can make faster by making Django and Vue work together like now with the profile info.

socialhome/frontend/components/streams/Stream.vue
```diff
@@ -96,16 +103,47 @@ export default Vue.component("stream", {
...
+
+            switch (this.$store.state.stream.name) {
+                case "followed":
+                    this.$store.dispatch(streamStoreOperations.getFollowedStream, options)
+                    break
+                case "public":
+                    this.$store.dispatch(streamStoreOperations.getPublicStream, options)
+                    break
+                case "tag":
+                    options.params.name = this.tag
```
jaywink, line 128: There must be some kind of problem in passing in `this.tag` since locally I accessed `http://127.0.0.1:8000/streams/tag/foo/` and got an error notification (those are nice btw! :)). The API call is this: `http://127.0.0.1:8000/api/streams/tag//` so tag is empty.

jaywink (changes requested):
A few regressions in local testing which need to be fixed. Otherwise looks great!

socialhome/frontend/stores/applicationStore.js
christophehenry: > I would like to re-introduce that so that the initial page load would add the first set of content and then Vue would handle all the fetching after that (load more, switching streams, etc).

So we would have to maintain to sets of templates. It doesn't look like a smart idea to me :/

christophehenry: Oh, yes, that's easy.

socialhome/frontend/components/streams/Stream.vue
christophehenry: Problem spotted. Mistype.

christophehenry:
Single content view is fixed as long as tags.

jaywink (approved):

jaywink:
Alright, tested locally, good to go, thanks! Federation suite tests are failing but as far as I can tell locally it's not on our side - certainly not this PR I think. So merging.

Issue #446, opened 2018-02-26
[API] /api-token-auth/ should return token and user ID


christophehenry:
Basic informations about the user should be made available for client from the API.

jaywink:
Closed via #447

PR #447, opened 2018-02-27
Add user profile at authentication

christophehenry:
`/api-token-auth/` now returns limited profile infos in addition to token:

```
{
  "guid": "eba9d938-71a2-4570-96c2-99e6e6bb232f",
  "handle": "karl_marx@127.0.0.1:8000",
  "home_url": "http:\/\/127.0.0.1:8000\/p\/eba9d938-71a2-4570-96c2-99e6e6bb232f\/",
  "id": 2,
  "image_url_small": "http:\/\/127.0.0.1:8000\/static\/images\/pony50.png",
  "is_local": true,
  "name": "",
  "url": "http:\/\/127.0.0.1:8000\/p\/eba9d938-71a2-4570-96c2-99e6e6bb232f\/",
  "token": "0ab231f9f7ab63a6f78fd9746b9bdc9926a734fd"
}
```

- [x] Tests
- [x] Update the docs to explain how to recover token

Refs: #446

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -443,6 +443,7 @@ Changed
 
 * Create and update content will now redirect to the content created or updated. Previous behaviour was user preferred landing page.
 * Delete content will now redirect back to the page where the delete was triggered from. Previous behaviour was user preferred landing page. If the content delete is triggered from the content detail page, redirect will happen to user preferred landing page as before. (`#204 <https://github.com/jaywink/socialhome/issues/204>`_)
+* ``/api-token-auth/`` endpoint now returns limited profile information in addition to token
 
 Fixed
 .....
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -41,6 +41,8 @@ Changed
 
   Since Vue streams all tweets are initialized programmatically as they are rendered in the stream so we don't need to have the script tag on each oembed separately.
 
+* ``/api-token-auth/`` endpoint now returns limited profile information in addition to token
+
 Fixed
 .....
 
@@ -443,7 +445,6 @@ Changed
 
 * Create and update content will now redirect to the content created or updated. Previous behaviour was user preferred landing page.
 * Delete content will now redirect back to the page where the delete was triggered from. Previous behaviour was user preferred landing page. If the content delete is triggered from the content detail page, redirect will happen to user preferred landing page as before. (`#204 <https://github.com/jaywink/socialhome/issues/204>`_)
-* ``/api-token-auth/`` endpoint now returns limited profile information in addition to token
 
 Fixed
 .....
--- a/socialhome/tests/test_viewsets.py
+++ b/socialhome/tests/test_viewsets.py
@@ -1,4 +1,5 @@
 from django.urls import reverse
+from rest_framework.authtoken.models import Token
 
 from socialhome.models import ImageUpload
 from socialhome.tests.utils import SocialhomeAPITestCase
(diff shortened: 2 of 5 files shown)


jaywink (commented):
Great! <3 Just needs a test since we're now owning the code from DRF and thus responsible for testing the view post doesn't crash and burn.

jaywink:
Oh yeah, the docs should also have a little update where they currently say how to retrieve the token to indicate it now returns also some profile info.

christophehenry:
Right. Added to the tasks.

christophehenry:
@jaywink I have a problem with [this test](https://github.com/jaywink/socialhome/pull/447/files#diff-67be71ea20afc07d2c4665880dc7fdb8R52). What's the way to create the user in DB so that Django REST API is aware of it?

Also, how can I write the Swagger documentation for this endpoint?

socialhome/tests/test_viewsets.py
```diff
@@ -25,3 +25,41 @@ def test_image_upload(self):
...
+        self.assertEqual(response.data["password"], ["This field is required."])
+
+    def test_unable_to_login(self):
+        response = self.client.post(reverse("api-token-auth"),
+                                    {"username": self.user.username, "password": "%sLol" % self.user.password})
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, {'non_field_errors': ['Unable to log in with provided credentials.']})
+
+    def test_authentication(self):
+        self.client.force_authenticate(self.user, token="0ab231f9f7ab63a6f78fd9746b9bdc9926a734fd")
```
jaywink: You're testing login so you shouldn't be forcing authentication in the test - this line shouldn't be here imho.

socialhome/tests/test_viewsets.py
```diff
@@ -25,3 +25,41 @@ def test_image_upload(self):
...
+        response = self.client.post(reverse("api-token-auth"),
```
jaywink: Since this is a `SocialhomeAPITestCase` you don't need to use `self.client`. Our test classes use some handy helper stuff which means you can just do `self.post("api-token-auth", {"username": self.user.username, "password": self.user.password})`. (see [here](https://django-test-plus.readthedocs.io/en/latest/usage.html#testing-drf-views))

Does that work? If not I'm not sure what could be the problem. The password should be set by the factory so that shouldn't be an issue..

jaywink:
> Also, how can I write the Swagger documentation for this endpoint?

I mean [this page](https://socialhome.readthedocs.io/en/latest/api.html#token-authentication) - but Swagger docs would be nice as well! They're done by adding a docstring to the view itself. See [this page](http://www.django-rest-framework.org/topics/documenting-your-api/#documenting-your-views). So basically, some generic title and a `post:` section should do it.

christophehenry:
Done. Ready for review.

docs/changelog.rst
```diff
@@ -443,6 +443,7 @@ Changed
 
 * Create and update content will now redirect to the content created or updated. Previous behaviour was user preferred landing page.
 * Delete content will now redirect back to the page where the delete was triggered from. Previous behaviour was user preferred landing page. If the content delete is triggered from the content detail page, redirect will happen to user preferred landing page as before. (`#204 <https://github.com/jaywink/socialhome/issues/204>`_)
+* ``/api-token-auth/`` endpoint now returns limited profile information in addition to token
```
jaywink: Changelog entry is in the wrong place, it should be in the latest upcoming release (top of file) not first release (bottom of file).

socialhome/views.py
```diff
@@ -49,3 +57,23 @@ def get_form_kwargs(self):
...
+    ])
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.serializer_class(data=request.data, context={"request": request})
+        serializer.is_valid(raise_exception=True)
+        user = serializer.validated_data["user"]
+        token, created = Token.objects.get_or_create(user=user)
+        data = LimitedProfileSerializer(user.profile, context={"request": self.request}).data
+        data.update({"token": token.key})
+        print(data)
```
jaywink: Print should be removed.

jaywink (commented):
A few minor comments, then good to go!

socialhome/views.py
christophehenry: Ah, missed this one after finishing to debug. Nice catch.

christophehenry:
Done.

jaywink (approved):
Great, thanks! <3

PR #471, opened 2018-03-27
First attempt at Dockerized production socialhome

ghost:
This is a small pull request, but there's a lot in this which we'll need to figure out.

The first thing to note is that we use a single image for all of the various Socialhome functions. That's not a normal Docker pattern, but due to the structure of Socialhome, the alternative would be to have a set of Docker images that are exactly the same except for the entrypoint. More on this a bit later.

Secondly, Docker Compose does not allow for multiple instances of a particular program. That is, it's not possible using Docker Compose to run multiple instances of RQScheduler. That leaves us with two alternatives: of a) Requiring Docker Swarm or b) Using something like Circus to manage these processes

While that's not supported in the current setup, my suggestion would be that for small instances, mirroring the Circus setup might be the easiest thing to do. Larger installations may want to use Docker Swarm or Kubernetes, but I'm hesitant to require it.

Thirdly, I've been giving a lot of thought and consideration to the users that these applications run as. Docker's use of Linux containers means that there should be no easy escalation path from root within a container to root outside of a container- that's the whole point of containers. That notwithstanding, I still think that privilege de-escalation is a good idea, the problem has to do with file permissions in the data volumes- that is between the mapping between the container's view of the world (its /etc/passwd) and the system's users, specifically mapping of UIDs and GIDs between the two. It would be fine (and easy) to run de-escalated within the container, but there can be problems with file permissions either between Docker images that don't share the same user, or between the Docker image and if (for some reason) that same area was shared with another system. The way this applies to Socialhome is in mainly regards to the media directory. I think the long term solution (before making this official) is to simply accept this limitation, but right now, everything runs as root.

Fourthly, I'm suggesting that a developer using this setup use nginx on an unencrypted port 80. The reason for this is that it likely makes sense that someone deploying this software will be using some kind of virtual host management and likely adding SSL/TLS encryption at that level. Therefore by offering the site unencrypted to that service, it can handle all the certificates itself, and or do caching.

Fifth, relatedly, I'm using the nginx official Docker image with a configuration file I supply myself. This involves an extra step for the user (deployer) in setting this up. We could offer our own image.

Sixth, I'm including an exim server for smart relay. I'm very much on the fence about this. On the argument for including it, this simplifies the Socialhome setup in the sense that the Socialhome software itself would only need to know about the "smtp" server, rather than needing to have the software talk to the actual mail server. This also eliminates the need for authentication credentials to the mail server to be shared to the various Python components, and also means that they don't need to even have network access to the final mail server- only the "smtp" server would need that. Lastly, it means that if the upsteam mail server is not up for some reason, it should allow for the mail sending process (one of the rqworker processes) to complete successfully.

On the 'con' side, there are a number of negatives. Firstly, this will add an extra level of complexity, and delay, to mail sending. Secondly, there's an argument that could be made that the notification system itself should handle the inability to send mail, rather than it completing successfully and then being unable to know if there was a failure later. Thirdly, I'm not entirely thrilled with this Docker image. It's unclear to me where Exim stores its mail queue data and without that, it would be possible that mail might be lost if the server is shut down. There are equivalent images for Postfix (which I'm more familiar with) but they seemed more complex to set up.

I almost went down of writing my own LSMTP server, and then decided that this was a very bad idea.

I don't consider this done and ready. I am giving this to you as something to discuss.

At the very least, before it's committed, it needs more testing and documentation.

Seventh and last, the production image should only contain the software needed to run, not all the software needed to build Socialhome. That means all the NPM build tools are unnecessary. There are a number of ways to do this, but I think the right way is simply to use a multi-stage build (the same way we might if we were compiling software). If we do this, we may also be able to take advantage of the opportunity to run `compileall` on the code, building .pyo and .pyc files to improve loading time.

Dovetailing this, though, we if we want to optimize this, we would also remove the javascript, as well as uwsgi, etc. from all but the necessary containers.

To highlight the most pressing questions:

1. Do we have one image with lots of entrypoints, or many images?

2. Handling the number of rqworkers required without adding more dependedencies

3. How do we want to handle users?

4. Do we want to ship instructions, or our own nginx image?

5. Do we want to suggest this exim container or maybe something else?

And remaining to do:

1. Testing this setup - I need to test it a few more times from scratch
2. Documentation - There's a lot to document, especially the boostrapping process
3. Deciding on some of the questions above
4. Putting up an official docker image for Socialhome


jaywink:
Great, thanks for working on this.

Just a few quick comments, need to get back to this many times to go through everything.

I'm ok with a single image running everything that is needed, using Circus to run everything. This would provide more simplicity. The point of having a docker possibility is to make deployment simple. For that reason, I would also prefer if there was a docker compose file to push up 1) the socialhome image (using circus + stuff inside), 2) postgres db and 3) redis. The latter two are an integral part of requirement to run, so we should provide a docker compose file to tie everything together.

In the simplest, the user should only have to set a few environment variables (socialhome hostname, secret key, etc must haves) and do `docker-compose start`. Anything else should be optional.

I'm ok with the exim mail server, but it should not imho be default. The problem with this is that email delivery is very important and without any extra setup email delivery will be flaky (ie going to spam). For good user experience I would recommend all admins to use a "proper" SMTP server unless they have the knowledge to setup their own - then they can configure the built in smtp server on their domain.

I wouldn't be too worried about node things or optimizing python - this is very micro optimization that can be handled later - if even makes sense. If you actually make benchmarks I doubt you're going to see much difference with the python compiled files and if you compare bytes saved on disk it wont be a lot compared to how much for example uploaded images will take :)

An example nginx/apache config for the reverse proxy would be great at some point :+1: 

Looking forward to see this develop!

docker/prod/.env.example
```diff
@@ -0,0 +1,14 @@
+DATABASE_URL=postgres://socialhome:socialhome@postgres:5432/socialhome
+
+DJANGO_SETTINGS_MODULE=config.settings.local
```
jaywink, line 3: This should be `config.settings.production`

docker/prod/Dockerfile.socialhome
```diff
@@ -0,0 +1,34 @@
+FROM python:3
+
+# Ensure pip and setuptools are up to date as well
+# We need a slightly older setuptools due to a bug in pip-tools
+RUN pip install -U pip setuptools==30.4 pip-tools
+
+## # We need Node for Javascript. Install it and then grab the latest version
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
```
jaywink, line 8: I'm pretty sure (at least a little while ago!) node 9 didn't work, some third party library breaks. If you're sure all the frontend compiles fine with node 9, we can change the relevant documentation then. Currently it says [6-8](https://socialhome.readthedocs.io/en/latest/install_guides.html#install-node-js) which I have personally tested to work.

docker/prod/Dockerfile.socialhome
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR /code
+RUN pip install --no-cache-dir -r ./requirements/requirements.txt
+
+COPY . .
+
+RUN cp .env.example .env
+RUN cp docker/prod/uwsgi.ini /code/uwsgi.ini
+# This file needs to be outside the /code dir as it's mounted  durin development
+RUN cp docker/prod/docker-entrypoint.sh.socialhome ./docker-entrypoint.sh
+
```
jaywink, line 27: I can't see any npm install / build and collectstatic steps?

docker/prod/docker-entrypoint.sh.socialhome
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+#
+# Socialhome Docker Production entrypoint
```
jaywink, line 3: This should be able to run circus if we go by that - no need to have uwsgi/daphne/runworker/rqworker/rqscheduler even available as options imho.

ghost:
I may be missing some of your reply, but let's see if we can address two non-trivial parts:

1. Circus

I think for small installations, this makes sense. The question of course is how much goes in Circus vs how much is outside. I think it makes sense to have all the Socialhome things in one service, which then just exposes uwsgi, daphne and the static files. Then we can ship an nginx server by default or people can use whatever they're comfortable with.

On the other hand, it might even make more sense to ship nginx inside this package then the whole thing is self-contained. I'm okay either way here.

Either way, I'd go one step further and say that this is probably a reasonable default- a single service people can start and stop called "socialhome" that uses circus under the covers.

I would hesitate to remove the other entrypoint options, in case someone does want to do something more complicated- we'll make it easy for them to see how it could be done. I'd also recommend somewhere (maybe just in docs or an example docker compose) showing how to decouple the services- since I think that's a useful exercise.

2. File Collection

You asked where the static and media file collection were, and this is one of the trickier parts of the bootstrap. I did remove them from the build, which was a mistake, but the trickier thing is that the frontend web server (nginx in this case) needs to access the static and media files (read only), while other services also need them (uwsgi and rqworker at the very least), so to do this, you need to share a data volume between the services.

This isn't a problem, but it's something that the user needs to be aware of. This dovetails slightly into the data volume ownership question/issue, but it's not a problem- just something to be aware of, just as it is with the Redis and Postgres images.

jaywink:
Bundling "socialhome" into one container was exactly what I was proposing, so I think this we have already resolved, to avoid using swarm, for now. But IMHO we need a ready to use docker compose to tie socialhome, redis and postgres together (and nginx if that is a container too, which is fine for me, sounds like a good idea). The compose file should pass all the required environment variables to the various containers, especially socialhome is meant to be configurable so many environment variables will need to be passed in. I would suggest passing a file through the compose file to the containers since we will be introducing new environment variables all the time and it would be bad to need to update the docker files every single time.

Using compose would ensure we can do the volumes correctly too without user having to worry about anything else than taking backups, right? I wouldn't worry about permissions that much at this point.

jaywink:
If I'm completely wrong about the need for a compose file please challenge me - I've only been running single containers in production myself. Running each one separately seems like a lot of effort and imho that is what compose is for, but I could be wrong and just need a practical example.

ghost:
Dcker best practices are still in evolution, and we're really discussing not the only Docker image Socialhome will ever have, but the official Docker image. Other people will be free to create their own Docker images as they wish.

The rest, let me explain to you why I understand your perspective, but disagree.

It would seem that your focus on this topic has shifted from it was on Matrix (IRC) from wanting to entirely decouple the software to now encapsulating the entire thing in one running Docker instance, probably to make it easier to deploy.

But the Docker view of the world is that of "services" An image runs a service and (usually) exposes that service on a port.  The question is then "Where does the service begin and end?" 

My source for an answer is to look at what other projects have done with their official images, so I'm going to show you two- Nextcloud and Wordpress. Both Nextcloud and Wordpress are "comparable" to Socialhome in that they're both applications that run as websites.

https://hub.docker.com/_/nextcloud/
https://hub.docker.com/_/wordpress/

If you look at both of these, you see that they both have instructions that include Docker Compose, and they both have a database as a dependency. The exception is that Nextcloud can run off an SQLite database instead. But of course, so could Socialhome.

You'll also see that both Wordpress and Nextcloud run a web server inside the image (Wordpress uses nginx and Nextcloud runs Apache). This is because they're both PHP applications and they need the web server to expose them as a running service.

For Socialhome, it's a bit different.

I still think encapsulating uwsgi, daphne and all of the other services into one image makes sense. But I don't think that the same logic extends to the databases (Redis and Postgres).

If you want other examples of the pattern I'm talking about, check out other images, including Teamspeak and Plone:

https://hub.docker.com/_/teamspeak/
https://hub.docker.com/_/plone/

jaywink:
I don't see any disagreement here. I was proposing the same thing, which IMHO you were proposing in the initial description, because going swarm way would be too difficult.

So, afaik what you are proposing, which I am agreeing to, is:

container 1:
* circus, which runs
  * daphne
  * uwsgi
  * rqworkers (configurable amount via env variable)
  * channels worker
  * rq schedyler

container 2:
* redis

container 3:
* postgres

container 4:
* nginx

To tie these together I suggested providing a docker compose file, which I can't see in this PR, unless it was left out by accident. We should provide an easy for to launch all containers tied together for those who want to do that, for example on a VPS. For advanced use cases people will use the dockerfile directly anyway. I would do that, I would make my own ansible role which uses the dockerfiles directly instead of using compose. I still think compose is a good idea to provide. We already have the dev version which can be modifed a bit.

The circus container is actually even following docker guidelines one could say. We're only running one process, circus ;)

ghost:
I'm wondering if there are some ideas from you or others on simplifying the installation.

I see two big problems, and I think neither is a huge deal but I'd like to find a solution to them.

The first is database role and database creation.

Since we're using Docker, the Postgres instance exists only for the specific instance of Socialhome. Even if someone runs two Socialhome instances, each instance would have its own entire Postgres server. Therefore I'm not as convinced that the Socialhome software needs to connect to Postgres as a specific "socialhome" user/role.  If we could either remove or automate this step, it would solve a lot of manual work that is necessary right now in the installation phase. An alternative to this would be for us to write a little script that we could call (maybe as part of manage.py) that simply created the role/user for us if it didn't exist, and also ran the migration for us. 

Similarly, gathering the static files and the media files has to be done inside of a data volume that's shared between the Socialhome software and the nginx instance. The way this is done now is that the user must start up a second instance of the Socialhome Docker image, then run the static collection in order to populate this data volume. It's a bit awkward. I'm wondering if instead, we should make this step part of the startup process? 
If we do this (both of these) then I think we'll have eliminated the more awkward steps in creating a Socialhome instance.

jaywink:
> The first is database role and database creation.

At least the official postgresql [image](https://hub.docker.com/_/postgres/) is able to create a db and role automatically via environment variables. I don't see whether it makes any difference what the role is called. As long as it is configurable for those who want to use their existing postgres server and not one with docker.

> The way this is done now is that the user must start up a second instance of the Socialhome Docker image, then run the static collection in order to populate this data volume. It's a bit awkward. I'm wondering if instead, we should make this step part of the startup process?

You should imho run static collection during the build step since the statics come from code, which doesn't change between runs, so doing it on run time would just make the container slow to start. Can't you just expose the path to the nginx container, somehow?

Media is different, there is no media shipped with code, and it's not collected. It's just a path shared between the two containers and shouldn't need any initialization or update commands.

Hope this helps!

ghost:
Yup.

I haven't had a chance to breathe this week but next week I should be able to finish this up.

There are other changes that I'm working on- like the intermediate build image (which builds the vue javascript). This addresses the complaint that we're shipping with node.

I'll post the newer version by Tuesday or Wednesday.

jaywink:
I don't really understand the complaint. Node is only required in the build step and people who use the images wont do the building - they can just pull a ready built image from docker hub? It's like complaining GCC was required for building a LibreOffice binary ;)

Maybe I just misunderstand the problem.

ghost:
The complaint is that when you get LibreOffice, you don't also get a copy of GCC. Our docker image is a "binary distribution" in that sense, and thus should not come with the build toolchain.

It's a legitimate complaint and will be relatively easy to correct.

jaywink:
I'm interested where this complaint comes from? I remember one discussion in the chatroom with someone complaining about it, but they were not complaining that a docker image would have node installed, they were complaining that *they need to install node* to run socialhome. The whole point of a docker container is that you don't need to know or care what it has to make it run.

I favour simple solutions over complex ones and would prefer not making the build process for the docker images unnecessarily complex - unless there is real need to do so. To justify the complexity, we should have clear benefit and I'm not sure what is real benefit of not having a `node_modules` directory and executable in the image - since in reality that is the only difference.

Of course if it's really simple to do I don't mind, but in general, simple is better than complex, since complex needs more maintenance time and has to justify the benefit.

jaywink:
Closing, see https://git.feneas.org/socialhome/socialhome/merge_requests/471#note_5983

PR #483, opened 2018-06-06
Add an easily customizable ``robots.txt`` with default rules

jaywink:
The rules by default disallow all except direct links to content, the
root profile and the public stream. Server admins can customize the
rules easily via the admin interface.


snake66:
Do I understand this right? You reveal that "all these URL endpoints exists, but please don't index them?" Not sure if that's a good idea if you really don't want anybody to prod these endpoints.

jaywink:
@snake66 many of these endpoints, especially `/streams/tag/<tagname/` is being heavily crawled and hit by bots currently. They arrive probably through some shared content and start following the tag links. As for the others, many of them are only available to authenticated users, so the robots file is just saying "no point in trying". If someone really wants to index the redirect to login form then for sure they can. For security aspect, all the endpoints are available in the code. If someone wants to hack Socialhome, they can find more information from the code - no point trying to hide the endpoints since we have to assume they are secure (and afaik they are :)).

We could just indicate `/` as default disallowed route to say to robots stay away from all pages, but I believe that would be a bad default. Search engines have a purpose -> to allow finding content. Was this what you had in mind?

snake66:
Most bots don't bother to read the code :) Some bots are nice and stays away as you've told them to, but a lot will use the robots.txt file as a starting point of where to prod. I don't doubt your endpoints are secure, but you may end up seing _more_ traffic to your endpoints, not less by this change. Would be interesting you are able to get some statistics for that.

PR #511, opened 2018-09-13
Api ui fix

### LINKED ISSUE #509: API UI not working
The docs mention a browsable API endpoint at `/api/` but both on my local development version and going to socialhome.network/api/ in the browser throws an error.

Having looked into Swagger a little bit, it looks like the problem is the `base.html` template was deleted in favor of having just `index.html` in Swagger v. 2.2.0. Testing locally, I found two possible solutions:
1. Deleting Socialhome's `rest_framework_swagger/index.html`. This allows the generated Swagger page to show up without customization.
2. Copying Swagger's `index.html` from the packages to the Socialhome templates and modifying it. This allows for customization, though is maybe less readable.

This is clearly a problem that rose from a change to Swagger, but should be easy enough to fix. If I know what fix is preferred I'd be happy to submit a PR.

ellisist:
Fixes #509 and keeps customization which was in place for the API browsable endpoint. Now `rest_framework_swagger/index.html` is overwritten directly, instead of extending (the no-longer-existent) `rest_framework_swagger/base.html`. This change was necessary because of a change in Swagger v 2.2.0. 
Also adds a very simple test which checks that the browsable endpoint `/api/` responds successfully. I put this in `config/tests/test_urls.py` (a new file) because `config/urls.py` is where Swagger is told to produce the API docs-- if there's a better place to put the test, let me know.

### DIFF
--- a/config/tests/test_urls.py
+++ b/config/tests/test_urls.py
@@ -1,0 +1,7 @@
+from socialhome.tests.utils import SocialhomeTestCase
+
+
+class TestSchemaView(SocialhomeTestCase):
+    def test_page_renders(self):
+        response = self.client.get("/api/")
+        self.assertEqual(response.status_code, 200)
--- a/socialhome/templates/rest_framework_swagger/index.html
+++ b/socialhome/templates/rest_framework_swagger/index.html
@@ -18,10 +18,9 @@
     <div class="topbar">
       <div class="wrapper">
         <div class="topbar-wrapper">
-          <a href="#" class="link">
-            <img src="{% static 'rest_framework_swagger/logo_small.png' %}" alt="Swagger Logo">
-            <span>swagger</span>
-          </a>
+          <div class="input" style="float: left; margin-top: 5px;">
+            <a href="/" class="header__btn">{{ request.site.name }}</a>
+          </div>
           <div class="download-url-wrapper">
           {% if USE_SESSION_AUTH %}
             {% if request.user.is_authenticated %}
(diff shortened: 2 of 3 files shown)


config/tests/test_urls.py
```diff
@@ -0,0 +1,7 @@
+from socialhome.tests.utils import SocialhomeTestCase
+
+
+class TestSchemaView(SocialhomeTestCase):
+    def test_page_renders(self):
+        response = self.client.get("/api/")
```
jaywink: Would prefer `reverse("api-docs")` instead of hardcoded url - if the route is changed in the URL configuration, then test wont need to be updated.

Also, the following is not mandatory to change, but as a tip. `SocialhomeTestCase` bases of `test_plus.TestCase` (https://django-test-plus.readthedocs.io/en/latest/), so you can shorten these two lines as follows:

```
self.get("api-docs")
self.response_200()
```

It's a great library for writing view tests :) And no need for reverse since test plus does that internally.

config/tests/test_urls.py
```diff
@@ -0,0 +1,7 @@
+from socialhome.tests.utils import SocialhomeTestCase
```
jaywink: I know this is the logical place for the test, but I feel still this file should be located in the root Socialhome app. Our project layout is maybe not optimal in this sense. So, I'd move this file to `socialhome/tests/test_api_docs.py`. It doesn't quite match the pattern, I see you've followed that closely, but I think the `config` module is maybe the wrong place for tests and code. Actually, due to this line in `pytest.ini`, this test is not even run by the CI: `testpaths = socialhome`

jaywink (changes requested):
Thanks! A few changes and then good to go!

jaywink (approved):
Thanks for the contribution! <3

PR #512, opened 2018-09-26
Use UUIDs for profile API

ellisist:
Changed the API to use UUIDs for profiles, and tests to test appropriately.
Wanted to have a sanity check before starting work on the rest of the API. 

TODO:
- [X] update frontend where necessary


jaywink (commented):
Awesome, thanks for the work!

Nice thing that you "fixed" the follow/unfollow API design. Previously I made it an  action _under_ the current user, thus the permission class is "IsOwnProfileOrReadOnly". This meant that one could only POST to the follow/unfollow actions if one owned the profile.

You have now switched it around so that the POST is made to the follow/unfollow action under the profile that the request user wants to follow. This makes a lot of sense and I totally agree we can change this, even though it is a major breaking change. Just need to adapt the frontend code for that.

I'm wondering does this change already work if you try in a REST client? I'm wondering if the permission class stops it working as the profile we're posting to is not "own or read only (GET)".

But all good to continue with this. Please also update any docstrings in the ProfileViewSet to match the endpoints new behaviour.

Also as a side note, to see more immediate progress, I'd recommend doing one API per PR instead of doing all at once. This will keep the PR smaller and be nicer for you also to get things merged. So you could for example finish the ProfileViewSet change here in backend and frontend first, then go to another API, adding the UUID in the model where needed. Content model as said already has it.

Ps. will try to get Travis tests to pass again, problems with SSL certs.

ellisist:
Frontend is updated and it seems to be working fine, there are four failures in the JS tests but it looks like it's not a problem with what I changed--the same tests fail on the master branch.

--EDIT--
There seems to be a bug, actually, where the profiles of socialhome users can't be found. It might be related to the permissions--the API to fetch a user only works on self for normal users, but works normally for a superuser.
(This doesn't impact following/unfollowing--just trying to view the profile of sh users)

jaywink:
I think to make the JS tests pass you need to fix the test data here: https://github.com/jaywink/socialhome/blob/86a7ce969da1d411f956e9dbeb52a58298fd6885/socialhome/frontend/tests/components/streams/Stream.tests.js#L60 and https://github.com/jaywink/socialhome/blob/86a7ce969da1d411f956e9dbeb52a58298fd6885/socialhome/frontend/tests/components/streams/Stream.tests.js#L104 .. ie assign an `uuid` to the fake profile in `applicationStore`.

ellisist:
Yeah it was the accounts set to private visibility. I wasn't expecting that to be the default, I guess.

The JS tests are passing now, let me know if it looks good. I didn't change the names of the endpoints yet, I agree it makes sense to do it in a separate PR.

jaywink (approved):
Looks great! Merging, thanks for the contribution and making the API more sane and final! <3

PR #513, opened 2018-10-16
Changing name of profile endpoints

ellisist:
Changed "add follower" to "follow" and "remove follower" to "unfollow". Also changed confirmation messages to match the change in name, updated tests appropriately.


jaywink (approved):
Looks great, nothing to add or remove! Thanks! :tada:

PR #514, opened 2018-10-17
UUID for stream API

ellisist:
Updated `profile-all` and `profile-pinned` endpoints to use UUIDs instead of ids, and changed tests appropriately!


socialhome/frontend/stores/streamStore.js
```diff
@@ -116,14 +116,14 @@ function newRestAPI(options) {
         })
         .get({
             action: streamStoreOperations.getProfileAll,
-            path: ({id, lastId = undefined}) => `${Urls["api-streams:profile-all"]({id})}${getLastIdParam(lastId)}`,
+            path: ({uuid, lastId = undefined}) => `${Urls["api-streams:profile-all"]({uuid})}${getLastIdParam(lastId)}`,
```
jaywink, line 119: Was trying to find where you change these being passed in but it seems that happens already in the previous PR ;)

jaywink (approved):
Good stuff! Thanks again :tada:

socialhome/frontend/stores/streamStore.js
ellisist: Yeah, I accidentally got ahead of myself I guess... And that actually broke fetching content from profiles, but it's fixed with this PR.

jaywink: Well I approved it so... :) For master branch it's "move fast break things".

